PR #94, opened 2014-07-01
ssh authentication and host key errors are unrecoverable

bdupras:
Perhaps this should re-raise because the transport can never become available in this condition.


jkeiser:
Closing this; #295 filed to track the grace period.

PR #98, opened 2014-07-05
merge current_machine_options into config in connect_to_machine

bdupras:
I'm not sure if this is the right thing to do.

In testing the machine_file resource with chef_metal_vsphere 0.3.0.beta.1, the config parameter passed to the driver's connect_to_machine does not contain a machine_options hash.  Without the machine_options, I'm not sure where to obtain ssh_config for the machine.

#am_i_doing_it_right

### DIFF
--- a/lib/chef_metal/chef_run_data.rb
+++ b/lib/chef_metal/chef_run_data.rb
@@ -63,7 +63,16 @@ module ChefMetal
       else
         machine_spec = ChefMetal::ChefMachineSpec.get(name, chef_server)
       end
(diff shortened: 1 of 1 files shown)


jkeiser:
I'm going to close this since the original patch isn't what we want to do.  The question of transferring passwords is open, but at the moment the answer is "transfer encrypted files, data bags, or any other secure mechanism you have; chef-provisioning will let you specify them in code on the other end, so you can use any storage Ruby can read."

PR #167, opened 2014-09-11
Fix issues when bootstrapping nodes within a proxy environment.

johnbellone:
When bootstrapping a node within a proxy environment we need this to
filter down to the Chef Client that is being installed/setup on the new node.

### DIFF
--- a/lib/chef_metal/convergence_strategy/precreate_chef_objects.rb
+++ b/lib/chef_metal/convergence_strategy/precreate_chef_objects.rb
@@ -5,12 +5,14 @@ require 'cheffish'
 module ChefMetal
   class ConvergenceStrategy
     class PrecreateChefObjects < ConvergenceStrategy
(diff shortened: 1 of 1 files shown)


jkeiser:
@dougireton @johnbellone note that you can now use this:

```ruby
machine_options convergence_options: {
  chef_config: <<-EOM
    http_proxy '...'
    https_proxy '...'
    no_proxy '...'
    # anything else you want in your chef config
  EOM
}
```

Given that, I think it's best to have people use `chef_config` rather than add specific variables for everything we want to configure.  If we need to do something automatic like pull the value of http_proxy from the environment then we should totally make a patch for that.  Please do reopen or say something if you have concerns or disagree; I have no problem with a continued conversation and there could be something I'm missing :)

vinyar:
@jkeiser do you feel the chef_config option is sufficiently documented in the readme?

`    chef_config: "log_level :debug\\n", # String containing additional text to inject into client.rb`
I kind of want to add a note with your example above to the main readme..

PR #254, opened 2014-12-23
Fix ssh_gateway values without a username in them

mivok:
This change fixes the behavior when an ssh_gateway is given that is simply a
hostname and doesn't include a username.

Previously, ssh_gateway values without an @ in them would result in the
hostname value being stored in gw_user and gw_host being nil, causing an
exception on the next line.

In addition, the line that sets a default username for the gateway to the same
as for the machine itself has been removed. This code never took effect anyway
due to the previous bug and it makes an incorrect assumption that the username
on the new machine is the same as the username on the gateway machine, which
is almost never the case given that the username on the new machine is
something like root, ec2user or ubuntu when creating a machine.

### DIFF
--- a/lib/chef/provisioning/transport/ssh.rb
+++ b/lib/chef/provisioning/transport/ssh.rb
@@ -280,8 +280,10 @@ module Provisioning
 
       def gateway
         gw_user, gw_host = options[:ssh_gateway].split('@')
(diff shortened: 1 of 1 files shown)


jkeiser:
Merged as 82cae87.  Thanks!

PR #255, opened 2014-12-23
Make bootstrap_proxy also set proxy in client.rb

mivok:
With knife, if bootstrap_proxy is set, then client.rb also contains relevant
http_proxy lines. This change makes chef-provisioning behave in the same way.


jkeiser:
Merged as 4fa8f91.  Thanks!

PR #256, opened 2014-12-24
Allow overriding ssl_verify_mode via convergence_options

mivok:
This lets someone change ssl_verify_mode in client.rb if they need to.


jkeiser:
Merged as b98715a (logic changed, but appears to still be correct).

PR #259, opened 2014-12-30
Correct Docker Driver Examples.

elliott-davis:
Updated require for docker driver.
Updated Apache requirement to be an actual cookbook in the supermarket.


jkeiser:
Merged as dfefa5e.  Thanks!!

PR #260, opened 2014-12-31
allow configurable log_level in msi install

andrewelizondo:
- mimic the install_sh convergence strategy to allow a configurable log level, cut the noise from `-l debug`


jkeiser:
Merged as e78541a.  Thanks!!!

PR #266, opened 2015-01-14
Updated the provisioning urls to default to ssl

xeon22:
Change the default urls for the chef install to https

### DIFF
--- a/lib/chef/provisioning/convergence_strategy/install_msi.rb
+++ b/lib/chef/provisioning/convergence_strategy/install_msi.rb
@@ -9,7 +9,7 @@ module Provisioning
 
       def initialize(convergence_options, config)
         super
-        @install_msi_url = convergence_options[:install_msi_url] || 'https://www.opscode.com/chef/install.msi'
+        @install_msi_url = convergence_options[:install_msi_url] || 'https://chef.io/chef/install.msi'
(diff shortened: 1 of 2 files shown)


lib/chef/provisioning/convergence_strategy/install_msi.rb
```diff
@@ -9,7 +9,7 @@ class InstallMsi < PrecreateChefObjects
 
       def initialize(convergence_options, config)
         super
-        @install_msi_url = convergence_options[:install_msi_url] || 'http://www.opscode.com/chef/install.msi'
+        @install_msi_url = convergence_options[:install_msi_url] || 'https://www.opscode.com/chef/install.msi'
```
jkeiser: Note to either you or us: these URLs should now be https://chef.io/chef/install.msi and https://chef.io/chef/install.sh.  Otherwise, this looks great!

jkeiser:
I'm on board with this completely, but not merging just yet--we need to do some testing.  I'm a little worried about its interaction with the recent Chef ssl https certificate verification stuff.

jkeiser:
Tested and committed as 414ee53 .  I tweaked the patch to add `www` back in since it was redirecting (and failing) without it.

Issue #263, opened 2015-01-13
NoMethodError: undefined method `gsub' for nil:NilClass following net-ssh patch upgrade


esciara:
As [discussed on gitter](https://gitter.im/opscode/chef-provisioning/archives/2015/01/13), a problem appeared following the patch `2.9.2` of the `net-ssh` gem. Symptoms are that the error below show when converging a `chef-provisioning` recipe (output in debug mode):
```
                      ================================================================================
                      Chef encountered an error attempting to load the node data for "simplest-server"
                      ================================================================================

                      Unexpected Error:
                      -----------------
                      NoMethodError: undefined method `gsub' for nil:NilClass

                      [2015-01-13T15:46:19+00:00] DEBUG: Re-raising exception: NoMethodError - undefined method `gsub' for nil:NilClass
                      /opt/chef/embedded/lib/ruby/gems/2.1.0/gems/mixlib-authentication-1.3.0/lib/mixlib/authentication/signedheaderauth.rb:120:in `canonical_path'
                        /opt/chef/embedded/lib/ruby/gems/2.1.0/gems/mixlib-authentication-1.3.0/lib/mixlib/authentication/signedheaderauth.rb:145:in `canonicalize_request'
                        /opt/chef/embedded/lib/ruby/gems/2.1.0/gems/mixlib-authentication-1.3.0/lib/mixlib/authentication/signedheaderauth.rb:93:in `sign'
                        /opt/chef/embedded/apps/chef/lib/chef/http/auth_credentials.rb:51:in `signature_headers'
                        /opt/chef/embedded/apps/chef/lib/chef/http/authenticator.rb:88:in `authentication_headers'
                        /opt/chef/embedded/apps/chef/lib/chef/http/authenticator.rb:43:in `handle_request'
                        /opt/chef/embedded/apps/chef/lib/chef/http.rb:222:in `block in apply_request_middleware'
                        /opt/chef/embedded/apps/chef/lib/chef/http.rb:220:in `each'
                        /opt/chef/embedded/apps/chef/lib/chef/http.rb:220:in `inject'
                        /opt/chef/embedded/apps/chef/lib/chef/http.rb:220:in `apply_request_middleware'
                        /opt/chef/embedded/apps/chef/lib/chef/http.rb:141:in `request'
                        /opt/chef/embedded/apps/chef/lib/chef/rest.rb:115:in `get'
                        /opt/chef/embedded/apps/chef/lib/chef/node.rb:496:in `load'
[...]

PR #267, opened 2015-01-15
Fix forward_port when using net-ssh 2.9.2.

causton81:
I did a bundle update today and got net-ssh 2.9.2 which causes a problem in forward_port. This patch fixes the problem for me, but I just noticed the whitespace is messed up :(

### DIFF
--- a/lib/chef/provisioning/transport/ssh.rb
+++ b/lib/chef/provisioning/transport/ssh.rb
@@ -307,15 +307,16 @@ module Provisioning
       end
 
       # Forwards a port over the connection, and returns the
-      def forward_port(local_port, local_host, remote_port, remote_host)
+      def forward_port(local_port, local_host, remote_port, requested_remote_host)
         # This bit is from the documentation.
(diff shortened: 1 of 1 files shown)


jkeiser:
Damn!  Let's see when this got introduced so we can do a version check.

jkeiser:
@causton81 can you explain the issue?  Your patch looks fine (though I'd like it if it either used the same `got_remote_host` pattern as `got_remote_port` or that you fix the `got_remote_port` pattern to match yours--either way) but I'm confused as to what started to go wrong.

jkeiser:
@lynchc pointed out that #263 is the actual issue here.

causton81:
Do you still want the variable name changes?

jkeiser:
@causton81 I made them locally and added a bunch of comments I should have had in there in the first place, retesting now :)

kramvan1:
I ran into this with ChefDK 0.3.6, using this patch seems to be working around it successfully.

PR #272, opened 2015-01-27
replace some getchefs with chef.ios; fix some copyright notices

smith:
I noticed that some things in the readme had "getchef.io" which wasn't a thing. Grepping came up with a few other instances of things that could be changed so here they are.


jkeiser:
@smith heh, I had the same OCD a few hours ago :)  It should all be fixed now!  Thanks :)

PR #275, opened 2015-02-05
Usability updates for my first time using chef-provisioning

tyler-ball:
\cc @jkeiser @fnichol

### DIFF
--- a/lib/chef/provider/machine.rb
+++ b/lib/chef/provider/machine.rb
@@ -19,11 +19,11 @@ class Machine < Chef::Provider::LWRPBase
   end
 
   action :allocate do
-    if current_driver && current_driver.driver_url != new_driver.driver_url
-      raise "Cannot move '#{machine_spec.name}' from #{current_driver.driver_url} to #{new_driver.driver_url}: machine moving is not supported.  Destroy and recreate."
-    end
     if !new_driver
       raise "Driver not specified for machine #{machine_spec.name}"
+    end
+    if current_driver && current_driver.driver_url != new_driver.driver_url
+      raise "Cannot move '#{machine_spec.name}' from #{current_driver.driver_url} to #{new_driver.driver_url}: machine moving is not supported.  Destroy and recreate."
     end
     new_driver.allocate_machine(action_handler, machine_spec, new_machine_options)
     machine_spec.save(action_handler)


lib/chef/provider/machine.rb
```diff
@@ -19,12 +19,12 @@ def whyrun_supported?
...
-    if current_driver && current_driver.driver_url != new_driver.driver_url
-      raise "Cannot move '#{machine_spec.name}' from #{current_driver.driver_url} to #{new_driver.driver_url}: machine moving is not supported.  Destroy and recreate."
-    end
     if !new_driver
       raise "Driver not specified for machine #{machine_spec.name}"
     end
+    if current_driver && current_driver.driver_url != new_driver.driver_url
+      raise "Cannot move '#{machine_spec.name}' from #{current_driver.driver_url} to #{new_driver.driver_url}: machine moving is not supported.  Destroy and recreate."
+    end
     new_driver.allocate_machine(action_handler, machine_spec, new_machine_options)
```
tyler-ball: This change means that if you remove your driver settings (and you are only using environmental variables) you will get the "Driver not specified for machine" message instead of a method being called on a nil object.

lib/chef/provisioning/chef_provider_action_handler.rb
```diff
@@ -42,16 +42,23 @@ def report_progress(description)
...
+      puts "\n"
     end
 
     def perform_action(description, &block)
       provider.converge_by(description, &block)
+      # The Chef event formatter normally handles all formatting, but short
+      # circuiting the resource/provider event chain by only calling `converge_by`
+      # skips this formatting logic.
+      # TODO this should definitely be refactored
+      puts "\n"
```
tyler-ball: This cleans up the STDOUT logging greatly when INFO level logs are enabled

lib/chef/provisioning/chef_provider_action_handler.rb
```diff
@@ -42,16 +42,23 @@ def report_progress(description)
...
         # We already did the action, but we trust whoever told us that they did it.
       end
+      puts "\n"
     end
 
     def performed_action(description)
       provider.converge_by description do
         # We already did the action, but we trust whoever told us that they did it.
       end
+      puts "\n"
```
jkeiser: Can we fix this better in converge_by?  What does this do?

jkeiser: Oh, is converge_by not filtering through the chef event formatter?  Probably the right thing is to make sure it does that.  That sucks :/

tyler-ball: `converge_by` is filtering through the chef event formatter just fine, but there are[other events](https://github.com/chef/chef/blob/master/lib/chef/formatters/doc.rb#L277-L281) that are NOT being called.  In the normal flow, `converge_by` only calls (resource_update_applied)[https://github.com/chef/chef/blob/master/lib/chef/formatters/doc.rb#L259-L274] via [why_run](https://github.com/chef/chef/blob/master/lib/chef/mixin/why_run.rb#L54).

If we have access to the `@events` bus from the run_context, we could call `resource_updated` ourselves.  But that feels pretty hacky to me - `provider.run_context.events.resource_updated(provider.new_resource, provider.action)`.

What about calling `provider.run_action` instead of `provider.converge_by`?  That would perform all the correct event formatting (I think...).

jkeiser: converge_by is supposed to be the right thing--we call it multiple times during the action, so run_action isn't the right thing.  Are you saying that resource_updated is not being called?

jkeiser: @tyler-ball `resource_updated` is only supposed to be called *after* all the converge_bys complete.  Like, typically, it goes:

```ruby
<start action>
resource_update_applied("set content of '/x/y.txt' to 'hello world")
resource_update_applied("set mode of '/x/y.txt' to 0666)
resource_updated(my_file_resource)
```

jkeiser:
@tyler-ball I think the core issue is that in Chef, logs don't go through the event formatter.  If they did, it would know that (updated) and (up to date) were in the midst of happening and could take appropriate action.  I think this is going to be a problem with literally any resource that does logging while taking an action.

lib/chef/provisioning/chef_provider_action_handler.rb
tyler-ball: @jkeiser I agree, `resource_updated` is only supposed to be called _after_ all the converge_bys complete.  If provider actions (the chef provisioning concept) only take place _inside_ providers then when the provider/resource pair exits, it will correctly un-indent the formatter and add a trailing newline.

So I think you are right - `converge_by` is the correct method to call, not `run_action`.

I would actually expect that core chef would have this same issue - `Chef::Log` lines mixed within event formatter lines would appear on the same line.  Aaaaaaaand I just read your comment below which says the same thing :smile: 

Some other ideas:

1. Force `Chef::Log` output to a file instead of stdout so it does not clutter the event formatter output.
1. Update the logging system in chef core to ensure log lines start at a new line, and if not, it adds one.

Thoughts?

jkeiser: I think we should revert the info log part of the fix and fix it in Chef.  It should be possible.  Maybe :)

tyler-ball:
I filed https://github.com/chef/chef/issues/2931 on the chef side to investigate fixing it over there.  As Dan says, the line is kept open so it can be appended with a `Up to Date` message.

Another idea I had is to subclass the Doc formatter and update the methods to have our own implementation.  Or we could add a new event to the system that only our formatter knows.

The other option is to always output log output to a file if a formatter is registered on STDOUT.  I think I like that idea best.

But regardless, I will clean up this PR and get it merged soon.

PR #279, opened 2015-02-09
Add machine.chef_config attribute to change client.rb

jkeiser:


lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb
```diff
@@ -177,18 +173,19 @@ def client_rb_content(chef_server_url, node_name)
...
+          chef_server_url #{chef_server_url.inspect}
+          node_name #{node_name.inspect}
+          client_key #{convergence_options[:client_pem_path].inspect}
+          ssl_verify_mode #{ssl_verify_mode.to_sym.inspect}
+        EOM
+        if convergence_options[:bootstrap_proxy]
+          content << <<-EOM
+            http_proxy #{convergence_options[:bootstrap_proxy].inspect}
+            https_proxy #{convergence_options[:bootstrap_proxy].inspect}
+          EOM
```
jtimberman: I think this is going to render the /etc/chef/client.rb with leading spaces.

jtimberman: Yup, it totally does:

```
          chef_server_url "https://api.opscode.com/organizations/jtimberman-chefconf"
          node_name "bootstrap-backend"
          client_key "/etc/chef/client.pem"
          ssl_verify_mode :verify_peer
```

Prior to this change these were column 1.

jtimberman:
Okay this totally works, it can set arbitrary configuration. As noted in the comments above we get some indentation on the lines. The fully rendered file looks like this:

```ruby
          chef_server_url "https://api.opscode.com/organizations/jtimberman-chefconf"
          node_name "bootstrap-backend"
          client_key "/etc/chef/client.pem"
          ssl_verify_mode :verify_peer
use_policyfile true
deployment_group 'chefconf-demo-provisioner'
```

The extra whitespace should be trimmed from the beginning of the lines in the heredoc, or they should be #gsub'd out, like is done in [test kitchen](https://github.com/test-kitchen/test-kitchen/blob/492104a266c87f65df2627b1f262b03c2d0f6fab/lib/kitchen/provisioner/chef_base.rb#L163-L171)

jtimberman:
One thing to note about this particular use case (using Policyfiles), the recipe attribute is ignored because policyfile's deployment group will take priority.

fnichol:
@jtimberman even better, I need to drop [this](https://github.com/test-kitchen/test-kitchen/blob/master/lib/kitchen/util.rb#L118-L134) in everywhere so it's more portable when refactoring heredoc blocks. After catching a particular RubyTapas episode I might even update the implementation :wink:

PR #286, opened 2015-02-20
Pass [] instead of nil when there are no machine specs

jkeiser:
Generally, drivers shouldn't have to worry about edge cases like nil machine_specs.  It just causes a multiplicity of bugs.


tyler-ball:
Makes sense, merged

PR #287, opened 2015-02-20
https_proxy environment variable for bootstrap install.

causton81:
install.sh now uses https URLs and the environment variable https_proxy needs to be set.


jkeiser:
Merged as 7a9fb53.

PR #290, opened 2015-02-22
Fallback to BSD md5(1)

### LINKED ISSUE #289: machine[].create idempotency checks fails in freebsd
The call in [1] doesn't work because the md5sum doesn't exist in FreeBSD.

I ran this against the Bento FreeBSD 10.1 vmware_fusion image.

[1] https://github.com/chef/chef-provisioning/blob/c57ff2d3ea1b7c19f4676dde99ee613ccab190c4/lib/chef/provisioning/machine/unix_machine.rb#L46

aespinosa:
Fixes #289

### DIFF
--- a/lib/chef/provisioning/machine/unix_machine.rb
+++ b/lib/chef/provisioning/machine/unix_machine.rb
@@ -44,6 +44,9 @@ module Provisioning
 
         # Get remote checksum of file
         result = transport.execute("md5sum -b #{path}", :read_only => true)
+        unless result.exitstatus == 0
+          result = transport.execute("md5 -r #{path}", :read_only => true)
+        end
         result.error!
         remote_sum = result.stdout.split(' ')[0]
 


jkeiser:
@aespinosa I'm worried about cross-shell compatibility here.  I think we may be safer if we check the return code from the first run of md5sum in ruby, and try "md5" if it's nonzero.

aespinosa:
Hi @jkeiser 

Ok, I'll make the shell check after reading the transport api a bit.

aespinosa:
@jkeiser How does this look now?

According to my tests on FreeBSD, exitstatus is 1 for csh and 127 for sh

aespinosa:
ping @jkeiser . let's hack on this tomorrow.

jkeiser:
Grar, I didn't get back to this as soon as I'd have liked :(  This looks good now, merging!

jkeiser:
Merged as 8b0e3a2

Issue #292, opened 2015-02-25
Don't sign requests as the provisioned client


jkeiser:
Due to this bug, we were forced to download and use the provisioned client's private key in order to fix ACLs on the server: https://github.com/chef/chef-server/issues/111

When that is fixed, please fix https://github.com/chef/chef-provisioning/blob/master/chef/provisioning/convergence_strategy/precreate_chef_objects.rb to not return the private key from `create_keys` or to use it in `grant_client_node_permissions`.

Saburesh07:
This issue or pull request is being automatically closed because the repository is being archived as part of the [Repository Standardization Initiative](https://github.com/chef-boneyard/oss-repo-standardization-2025).

PR #293, opened 2015-02-25
Fix #59: set node permissions correctly before converging

### LINKED ISSUE #59: Initial chef-client run on workstation talking to hosted chef-server fails on creating client
I'm using the 0.11.beta2 version of chef-metal. 

When I run the initial chef-client run on my workstation to create the machine, it seems to fail when it trys to create the client object on the hosted web server.

I'm guessing that my workstation chef-client doesn't have the permissions to create a new client?

Here's a gist with the last part of the chef-client run output:
https://gist.github.com/rberger/d4a37130dfd2f2869eec

Here's a gist with the stacktrace:
https://gist.github.com/rberger/745bc65fc803b6d91402

jkeiser:
Turns out we have the right code, but the server has a bug that cripples many installations.  Bug #292 has more detail on the workaround and issue.

### DIFF
--- a/lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb
+++ b/lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb
@@ -12,9 +12,9 @@ module Provisioning
 
       def setup_convergence(action_handler, machine)
         # Create keys on machine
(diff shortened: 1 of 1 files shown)


fnichol:
Too bad it came to this, but at least there's a way to move forward. :+1:

PR #294, opened 2015-02-26
Don't save after converge (that destroys attributes created by the converge)

### LINKED ISSUE #209: chef-metal is nuking 'normal' attributes on every converge.
According to https://docs.getchef.com/essentials_cookbook_attribute_files.html 'normal' attrbutes are supposed to be persisted in the node object. However, chef-metal (or possibly chef-zero) seems to be nuking them on every converge, causing mayham in the development of a cookbook.

jkeiser:
Fixed #209

### DIFF
--- a/lib/chef/provider/machine_batch.rb
+++ b/lib/chef/provider/machine_batch.rb
@@ -54,15 +54,12 @@ class MachineBatch < Chef::Provider::LWRPBase
       if m[:resource] && m[:resource].converge
         Chef::Log.info("Converging #{m[:spec].name} because 'converge true' is set ...")
         m[:machine].converge(m[:action_handler])
-        m[:spec].save(m[:action_handler])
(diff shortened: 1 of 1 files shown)


tyler-ball:
This is awesome that it fixes a nasty bug, but why was it there in the first place?  What was the reasoning?

jkeiser:
@tyler-ball there is no purpose to it: the only reason we ever save a machine_spec is to save off the information that user specified in the resource and the instance information.  It just shouldn't be done after setup: converges don't change the instance location or the user-specified data.  And the bug here is that if the converge adds information to the node, we overwrite it with our old (stale) copy.

This is giving machine_batch parity with machine, which does the same: https://github.com/chef/chef-provisioning/blob/master/lib/chef/provider/machine.rb#L53-L68

PR #297, opened 2015-02-27
Create generic "spec_registry" which will allow drivers to

jkeiser:
easily persist and retrieve registries of their items (such as aws_subnet, etc.).
Convert machine, machine_image and load_balancer to use it.

We should talk about it :)  Features will go into the AWS driver.  This is a backcompat breaking change for machine_images, because we're no longer storing machine_options in the data bag (which we never should have in the first place).

This particularly provides an external place for resources and drivers to grab registered locations of things like subnets, so that they don't have to funnel through resource (which is weird).

I haven't tested every single thing yet (will finish that tomorrow), but it seems to be working well for machines and for other things like aws_subnet.


lib/chef/provider/load_balancer.rb
```diff
@@ -20,13 +19,16 @@ def new_driver
...
+
       action :create do
-        lb_spec = Chef::Provisioning::ChefLoadBalancerSpec.get(new_resource.name) ||
-                  Chef::Provisioning::ChefLoadBalancerSpec.empty(new_resource.name)
+        lb_spec = chef_managed_entry_store.get_or_new(:load_balancer, new_resource.name)
 
         Chef::Log.debug "Creating load balancer: #{new_resource.name}; loaded #{lb_spec.inspect}"
         if new_resource.machines
-          machine_specs = new_resource.machines.map { |machine| get_machine_spec(machine) }
+          machine_specs = new_resource.machines.map { |machine| get_machine_spec!(machine) }
```
fnichol: Does the `!` on `#get_machine_spec!` indicate a side effect--curious about the name change

lib/chef/provider/machine_image.rb
```diff
@@ -20,49 +19,73 @@ def new_driver
     @new_driver ||= run_context.chef_provisioning.driver_for(new_resource.driver)
   end
 
+  def chef_managed_entry_store
+    @chef_managed_entry_store ||= Provisioning.chef_managed_entry_store(new_resource.chef_server)
+  end
```
fnichol: Could there be a common superclass or module mixin hiding with the `#chef_managed_entry_store` methods in the providers? I'm not sure either way, but at least the impls are the same--which is relieving

lib/chef/provisioning/chef_run_data.rb
```diff
@@ -86,10 +85,10 @@ def driver_for(driver)
     end
 
     def connect_to_machine(name, chef_server = nil)
-      if name.is_a?(MachineSpec)
+      if name.is_a?(ManagedEntry)
         machine_spec = name
       else
-        machine_spec = Chef::Provisioning::ChefMachineSpec.get(name, chef_server)
+        machine_spec = Provisioning.chef_managed_entry_store(chef_server).get(:machine, name)
       end
```
fnichol: Backwards compatibility I'm guessing? Woot!

lib/chef/provisioning/driver.rb
```diff
@@ -102,11 +102,11 @@ def driver_options
     # gears moving, but the job doesn't need to be done :)
     #
     # @param [Chef::Provisioning::ActionHandler] action_handler The action_handler object that is calling this method
-    # @param [Chef::Provisioning::MachineSpec] machine_spec A machine specification representing this machine.
+    # @param [Chef::Provisioning::ManagedEntry] machine_spec A machine specification representing this machine.
     # @param [Hash] machine_options A set of options representing the desired options when
```
fnichol: Yep, ManagedEntry is growing on me looking at the changes. Feels like I can understand it better than MachineSpec

lib/chef/provisioning/load_balancer_spec.rb
```diff
@@ -1,79 +1,14 @@
...
-    # could be lost, you should do this as quickly as possible.  Data will be
-    # saved automatically for you after allocate_machine and ready_machine.
-    #
-    def save(action_handler)
-      raise "save unimplemented"
-    end
+  class LoadBalancerSpec < ManagedEntry
+    alias :location :reference
+    alias :location= :reference=
   end
```
fnichol: Here's the payoff, nice and tidy!

lib/chef/provisioning/machine_spec.rb
```diff
@@ -39,43 +39,26 @@ def name
     #
     # chef-provisioning will do its darnedest to not lose this information.
     #
-    def location
-      chef_provisioning_attr('location')
+    def reference
+      attrs['reference'] || attrs['location']
```
fnichol: Would a falsey (i.e. `nil`, `false`) value for `attrs['reference']` lead to issues here? If so another way could be using fetch: `attrs.fetch('reference') { attrs['location'] }` (lazy block version used)

lib/chef/provisioning/managed_entry.rb
```diff
@@ -0,0 +1,121 @@
...
+    #
+    # This MUST include a 'driver_url' attribute with the driver's URL in it.
+    #
+    # chef-provisioning will do its darnedest to not lose this information.
+    #
+    def reference
+      # Backcompat: old data bags didn't have the "reference" field.  If we have
+      # no reference field in the data, and the data bag is non-empty, return
+      # the root of the data bag.
+      attrs['reference'] || attrs['location'] || (attrs == {} ? nil : attrs)
```
fnichol: Same `nil` value question as above. Again, it may not be an issue, it depends on the potential and actual values allowed in `attrs`.

lib/chef/provisioning/managed_entry_store.rb
```diff
@@ -0,0 +1,136 @@
...
+      #
+      # Get the given data
+      #
+      # @param resource_type [Symbol] The type of thing to retrieve (:machine, :machine_image, :load_balancer, :aws_vpc, :aws_subnet, ...)
+      # @param name [String] The unique identifier of the thing to retrieve
+      #
+      # @return [Hash,Array] The data.  Will be JSON- and YAML-compatible (Hash, Array, String, Integer, Boolean, Nil)
+      #
+      def get_data(resource_type, name)
+        raise NotImplementedError, :get_data
```
fnichol: Nice choice, I like `NotImplementedError` :smile:

fnichol:
Comments inline above, but nothing critical. In general, I'm :+1:

fnichol:
And to be clear, I mean merge if ready!

lib/chef/provider/load_balancer.rb
jkeiser: There are two versions: one that will error out if the machine spec isn't there (requires it to exist), and the other will return nil.

lib/chef/provider/machine_image.rb
jkeiser: There could be--I'm trying to avoid even further change in this one PR :)

lib/chef/provisioning/chef_run_data.rb
jkeiser: Indeed!

lib/chef/provisioning/machine_spec.rb
jkeiser: `nil` and `false` aren't valid values for it anyway, making this a bit more readable--but yes, I loves me some `fetch`.

PR #302, opened 2015-03-03
Add show_output attribute for machine_execute.

causton81:
I have a `machine_execute` at the end of my provision recipe that runs system integration tests from a node in the cluster. It's nice to see the output without having to use `-l debug`. I use chef-provisioning as a developer so that my co-workers can provision their own development clusters and it's cool to run the tests at the end automatically.

I suppose a similar effect could be achieved by capturing the output on the remote server and then downloading and displaying it on the provisioning server...

### DIFF
--- a/lib/chef/provider/machine_execute.rb
+++ b/lib/chef/provider/machine_execute.rb
@@ -27,7 +27,7 @@ class MachineExecute < Chef::Provider::LWRPBase
   end
 
   action :run do
-    machine.execute(action_handler, new_resource.command)
+    machine.execute(action_handler, new_resource.command, :stream => new_resource.show_output)
   end
 
 end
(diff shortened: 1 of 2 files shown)


jkeiser:
Hmm, it might be a good idea to mimick the arguments to `machine.execute` here--what do you think about calling it `stream`?  I'm a huge :+1: on having an attribute :)

causton81:
I think `stream` describes how it actually shows the output as opposed to some sort of buffering, so stream is a better name. :+1:

jkeiser:
Renamed to `live_stream` to better match the Chef ShellOut stuff, and checked in as d99ad1557a1f15fea66e378750b34fc6b5601fd3!  Thanks :)

causton81:
Awesome!

PR #303, opened 2015-03-04
Install/Upgrade if the chef-client (or the specified version) does not exist

bbbco:
Chef installer takes care of upgrade/downgrade automagically!

Please note, that we could definitely improve the error handling if a bad version is passed, as the chef error is pretty vague.

### DIFF
--- a/lib/chef/provisioning/convergence_strategy/install_cached.rb
+++ b/lib/chef/provisioning/convergence_strategy/install_cached.rb
@@ -42,14 +42,15 @@ module Provisioning
       def setup_convergence(action_handler, machine)
         super
 
-        # Install chef-client.  TODO check and update version if not latest / not desired
-        if machine.execute_always('chef-client -v').exitstatus != 0
-          platform, platform_version, machine_architecture = machine.detect_os(action_handler)
-          package_file = download_package_for_platform(action_handler, machine, platform, platform_version, machine_architecture)
-          remote_package_file = "#{@tmp_dir}/#{File.basename(package_file)}"
-          machine.upload_file(action_handler, package_file, remote_package_file)
-          install_package(action_handler, machine, remote_package_file)
-        end
+        # Install chef-client.
+        version = machine.execute_always('chef-client -v')
+        return if version.exitstatus == 0 and version.stdout().strip().include?(" #{chef_version}")
+
+        platform, platform_version, machine_architecture = machine.detect_os(action_handler)
+        package_file = download_package_for_platform(action_handler, machine, platform, platform_version, machine_architecture)
+        remote_package_file = "#{@tmp_dir}/#{File.basename(package_file)}"
+        machine.upload_file(action_handler, package_file, remote_package_file)
+        install_package(action_handler, machine, remote_package_file)
       end
 
       def converge(action_handler, machine)
--- a/lib/chef/provisioning/convergence_strategy/install_cached.rb
+++ b/lib/chef/provisioning/convergence_strategy/install_cached.rb
@@ -51,7 +51,7 @@ module Provisioning
         if version.exitstatus == 0
           if !chef_version
             return
-          elsif version.stdout.strip =~ /Chef: #{chef_version}([^0-9]|$)/
+          elsif version.stdout.strip =~ /Chef: #{chef_version}$/
             return
           end
         end
--- a/lib/chef/provisioning/convergence_strategy/install_cached.rb
+++ b/lib/chef/provisioning/convergence_strategy/install_cached.rb
@@ -45,10 +45,13 @@ module Provisioning
         # Check for existing chef client.
         version = machine.execute_always('chef-client -v')
 
+        # Don't do install/upgrade if a chef client exists and
+        # no chef version is defined by user configs or
+        # the chef client's version already matches user config
         if version.exitstatus == 0
-          if chef_version and version.stdout.strip =~ /Chef: #{chef_version}([^0-9]|$)/
+          if !chef_version
             return
-          elsif !chef_version
+          elsif version.stdout.strip =~ /Chef: #{chef_version}([^0-9]|$)/
             return
           end
         end
--- a/lib/chef/provisioning/convergence_strategy/install_cached.rb
+++ b/lib/chef/provisioning/convergence_strategy/install_cached.rb
@@ -42,10 +42,18 @@ module Provisioning
       def setup_convergence(action_handler, machine)
         super
 
-        # Install chef-client.
+        # Check for existing chef client.
         version = machine.execute_always('chef-client -v')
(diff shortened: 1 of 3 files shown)


bbbco:
Also, note that the Chef Installer takes care of handling the various numberings of versions (i.e. "12" vs "12.1" vs "12.1.0") on its own.

lib/chef/provisioning/convergence_strategy/install_cached.rb
```diff
@@ -42,14 +42,15 @@ def initialize(convergence_options, config)
...
-          platform, platform_version, machine_architecture = machine.detect_os(action_handler)
-          package_file = download_package_for_platform(action_handler, machine, platform, platform_version, machine_architecture)
-          remote_package_file = "#{@tmp_dir}/#{File.basename(package_file)}"
-          machine.upload_file(action_handler, package_file, remote_package_file)
-          install_package(action_handler, machine, remote_package_file)
-        end
+        # Install chef-client.
+        version = machine.execute_always('chef-client -v')
+        return if version.exitstatus == 0 and version.stdout().strip().include?(" #{chef_version}")
+
```
jkeiser: I know it's a small thing, but it's easy to miss the `return if`.  Can you either put a comment before `return if` saying `# if we already have the right version, don't install / upgrade`, or re-wrap it in the `if`?  Otherwise, the change looks good :)

jkeiser:
@bbbco just one super minor nit that applies to all three files, otherwise :+1: and thanks!

jkeiser:
Actually, going through it I found two more potential issues:

1. If you don't specify a version (a common case), it probably shouldn't check anything (might be harmless, but intent is muddy).
2. If you specify a 1-digit or 2-digit version, the regex may mess up: specify "1" and it will match "11."

bbbco:
Thanks @jkeiser . In response:

1. Sure, I can add some logic no problem

2. The chef installer just figures out the version to select based on whatever is passed in. The same thing happens if you pass "1" to the installer script itself. Should we really do extra parsing to try to figure out intent? Shouldn't this be an issue we pass onto the core Chef team?

jkeiser:
The issue is with our parsing:

If the Chef Client returns "Chef Client version 11.0.1", and the user asks us for version "1", we will decide that he already has version 1 and do nothing.

bbbco:
Yeah, sorry, I get what you're saying. Have a better suggestion? Would be nice to be able to support "11" vs "11.0" vs "11.0.1"

Maybe something like:
````
current_version = Gem::Version.new(version.stdout().strip().split(" ")[-1])
desired_version = Gem::Version.new(chef_version)
if current_version == desired_version
    return
````

**Comment by jkeiser** [comment]
Maybe `=~ /#{version}([^0-9]|$)/`?  That would make sure you didn't match extra digits.

**Comment by chef-supermarket** [comment]

* @bbbco

**Comment by chef-supermarket** [comment]

**Comment by bbbco** [comment]
Updated with fixes suggested by @jkeiser

**Comment by bbbco** [comment]
BTW, I had signed the Chef CLA, but just not linked my account to the Chef Supermarket, which apparently will trigger the above bot ... go figure

**Comment by jkeiser** [comment]
:+1: but I'm not in a place I can merge this week cc @tyler-ball

File: lib/chef/provisioning/convergence_strategy/install_cached.rb [hunk 2]
```diff
@@ -42,14 +42,26 @@ def initialize(convergence_options, config)
...
+        # Check for existing chef client.
+        version = machine.execute_always('chef-client -v')
+
+        # Don't do install/upgrade if a chef client exists and
+        # no chef version is defined by user configs or
+        # the chef client's version already matches user config
+        if version.exitstatus == 0
+          if !chef_version
+            return
+          elsif version.stdout.strip =~ /Chef: #{chef_version}([^0-9]|$)/
```
tyler-ball: This wouldn't work in the case where you are trying to install version `12.1.0` and you already have installed `12.1.0.dev.0`. I think we should just try and do an exact match because the script will handle the case where you specify `12` and `12.1.0` is already installed.  Otherwise we are going to start duplicating that logic here.  So `/Chef: #{chef_version}$/`

jkeiser: The thing we need to make sure of is that we don't print green text (and mark the resource updated) of the script decides no upgrade needs to be performed. That will take some doing, but I'd be happy to help next week. Meantime, do you feel we should take the patch and file a bug for that case, try to fix first, or wait?

tyler-ball: I think the only way to prevent updating the resource every run is to specify the exact version of chef that is desired, or for us to duplicate logic from the install.sh script.  If the user provides `12.1` as the desired chef version, `12.1.0` is the currently installed version and `12.1.1` was just released, what should we do?  Either we call the install.sh script and let it handle updating, or we duplicate the logic in install.sh so we can know whether we _should_ update.

I'm now realizing we may have to duplicate that logic regardless.  install.sh is only 1 of the 3 convergence strategies here.  The other 2 won't be able to answer the above question unless we access the omnitruck API.

My vote is that we change it to exact matching for all 3 convergence strategies.  This will make it idempotent for users who specify the full version of chef they want to install.  Users who specify a partial version will get the existing behavior, so we should file a bug to come back and fix that.  Sound good?

bbbco:
I deferred to @tyler-ball's wisdom, and updated the PR with logic that checks against the exact version. If someone wants to attempt to create a parser to figure out the versions in a more accurate way, they can :)

jkeiser:
Either way, we need to fix any issues with green text / updating.  Non-convergent resources are a non-starter and will cause re-converge and all sorts of other unnecessary things.

@tyler-ball it's entirely possible to run the script every single time, look at the script output, and *retroactively* print the green text / set the resource update flag afterwards.  That's what we should do.

jkeiser:
Specifically, "running the script" is not necessarily a wrong thing to do every run.  It does the test-and-set for us.

jkeiser:
On the other hand, we *cannot* run the script in why-run mode unless the script itself has a why-run (which I think it might).

jkeiser:
We should not block this checkin on it--I think exact matching will be perfectly fine as long as we are super clear about the requirement.  Is there a way we can tell the user if they have specified a version that will never match?  Like 2 vs. 2.0 vs. 2.0.0?  Or will they all match?

tyler-ball:
No, if the user specifies 2 and version 2.0.0 is already installed, it will fall through to the script every time.  I'll add some logic to compare the supplied version to the current version, and output a message telling to the user to specify tighter version constraints if they don't want the resource converging every time.

tyler-ball:
Merged as https://github.com/chef/chef-provisioning/commit/37b53d557af19f8b8947f802d48960849f39e134

PR #312, opened 2015-03-26
Use the actual `name` from the superclass, else we get caught in a loop

jkeiser:


jkeiser:
CC @tyler-ball @fnichol

PR #314, opened 2015-04-03
Update chef gem to fix version conflict in ChefDK

teknofire:
This is an update to the readme with instructions on how to fix the version conflict problem with ChefDK 0.4.0 locking chef-zero to 3.2.1 which conflicts with the requirement for version ~4.0 of the gem


jkeiser:
Sorry for the late merge.  post-ChefConf craziness.

PR #317, opened 2015-04-08
Install chef-client using Proxy

afiune:
As a Enterprise User I need to be able to tell chef-provisioning to use my corporate proxy so I can install chef client on my enterprise.

cc/ @jonsmorrow @mfdii @jkeiser @smford22

### DIFF
--- a/lib/chef/provisioning/convergence_strategy/install_sh.rb
+++ b/lib/chef/provisioning/convergence_strategy/install_sh.rb
@@ -64,17 +64,19 @@ module Provisioning
 
         # Install chef client
         # TODO ssh verification of install.sh before running arbtrary code would be nice?
-        if bootstrap_env.empty?
+        if convergence_options[:bootstrap_proxy].empty?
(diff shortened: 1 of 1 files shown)


afiune:
@jkeiser any update on this PR?

lib/chef/provisioning/convergence_strategy/install_sh.rb
```diff
@@ -64,7 +64,18 @@ def setup_convergence(action_handler, machine)
 
         # Install chef client
         # TODO ssh verification of install.sh before running arbtrary code would be nice?
-        @@install_sh_cache[install_sh_url] ||= Net::HTTP.get(URI(install_sh_url))
+        if bootstrap_env.empty?
```
jkeiser: @afiune any chance you can do `if convergence_options[:bootstrap_proxy]` instead?  It took me a while to grok what was going on since bootstrap_env isn't used in the if or else :)

Other than that, :+1:

PR #332, opened 2015-04-24
bumping winrm dependency to 1.3.0

mwrock:
bumping winrm dependency to avoid conflicts when using provisioner as a test-kitchen driver with new test-kitchen winrm functionality

### DIFF
--- a/lib/chef/provisioning/transport/winrm.rb
+++ b/lib/chef/provisioning/transport/winrm.rb
@@ -86,12 +86,12 @@ $file.Close
         # If you can't pwd within 10 seconds, you can't pwd
         execute('pwd', :timeout => 10)
         true
-      rescue Timeout::Error, Errno::EHOSTUNREACH, Errno::ETIMEDOUT, Errno::ECONNREFUSED, Errno::ECONNRESET, ::WinRM::WinRMHTTPTransportError, ::WinRM::WinRMWebServiceError, ::WinRM::WinRMWSManFault
-        Chef::Log.debug("unavailable: network connection failed or broke: #{$!.inspect}")
-        disconnect
-        false
       rescue ::WinRM::WinRMAuthorizationError
         Chef::Log.debug("unavailable: winrm authentication error: #{$!.inspect} ")
+        disconnect
+        false
+      rescue Timeout::Error, Errno::EHOSTUNREACH, Errno::ETIMEDOUT, Errno::ECONNREFUSED, Errno::ECONNRESET, ::WinRM::WinRMError
+        Chef::Log.debug("unavailable: network connection failed or broke: #{$!.inspect}")
         disconnect
         false
       end
--- a/chef-provisioning.gemspec
+++ b/chef-provisioning.gemspec
@@ -18,7 +18,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'net-ssh-gateway', '~> 1.2.0'
   s.add_dependency 'inifile', '~> 2.0'
   s.add_dependency 'cheffish', '~> 1.1'
-  s.add_dependency 'winrm', '~> 1.2.0'
+  s.add_dependency 'winrm', '~> 1.3'
 
   s.add_development_dependency 'rspec'
   s.add_development_dependency 'rake'


jkeiser:
Is there any reason we shouldn't just change it to `~> 1.2`?  That would be less risky with respect to other projects and allow the test-kitchen dep.

mwrock:
ah of course. 'm assuming you mean `>= 1.2` and not `~>`?

mwrock:
Just found a breaking change from 1.2.0 to 1.3.0 in winrm (see my last commit to this pr). The `WinRMWebServiceError` was removed in 1.3 which chef-provisioning rescues and thus gets a TypeLoadExceptioon. 1.3 adds a `WinRMError` as a base class for all winrm errors but rescuing that would require a hard 1.3 dependency. What we could do is rescue `StandardError` which is generally a bad practice. We can still weed out auth exceptions and use a mode general error message for the wider rescue.  I'm unsure how to not either break 1.3 or 1.2 without doing that.

Thoughts?

jkeiser:
@mwrock I guess we could detect the lack of `WinRMError` and rescue based on that :/  Like,

```ruby
rescue defined?(WinRMError) ? WinRMError : StandardError
```

danielsdeleo:
What's the risk in bumping to a dep on 1.3 (curious about the details). In any case, there's a lot of ways to work around it if you need to. Also winrm 1.3 removes nokogiri as a dependency which I'm really keen on.

jkeiser:
@danielsdeleo the risk is just that our code breaks :)  We try to catch their base error type, which no longer exists.  Have to fix that, then we can upgrade.

tyler-ball:
What is the issue with adding `'winrm', '~> 1.3'` in the gemfile?  Then we could count on `WinRMError` existing.  Is that going to conflict with something else in the omnibus, like Test Kitchen?  If so, we should just update Test Kitchen too.

mwrock:
The concern is if someone has a gem conflict with something else requiring `'~> 1.2` and they might be pulling down chef-provisioning a-la-carte separately from an omnibus install. This might be more common now with chef-provisioning since until recently that has been the standard for consuming it.

tyler-ball:
But if someone else requires `'~> 1.2'` then version 1.3 still satisfies that.  It is only an issue if someone requires `'~> 1.2.0'`.  My vote is that we test the hard dependency on winrm 1.3.0 in the latest ChefDK.  If that works then we :shipit:.  If it doesn't, we see if we can update the conflicting dependency parent and ship ChefDK 0.5.2.  Thoughts?

My thought is that we're trying to ensure that the ChefDK is the standard development environment, so if it works there we're good.  If someone has a provisioning node with only Chef installed (maybe they're installing provisioning via a recipe) and they have a hard dependency on WinRM 1.2 then we can help them upgrade, or they can stick with an earlier version of Chef Provisioning.

danielsdeleo:
I'm :+1: on this plan

danielsdeleo:
BTW, looks like knife-windows already upgraded: https://github.com/chef/knife-windows/blob/master/knife-windows.gemspec#L17

mwrock:
I'm really sorry and I think that maybe I have a fundamental misunderstanding of ruby versioning based on a recent diff between this thread and my current knowledge. I have been under the understanding that '~> 1.2` is the same as '~> 1.2.0' both meaning that 1.2.0 - 1.2.xx is an acceptable version. So that being the case `~> 1.2` would conflict with 1.3. It sounds like you are talking about `>= 1.2` rather than `~> 1.2`.

Sorry for the noob question but what exactly dos the '~> 1.2` constraint allow?

danielsdeleo:
@mwrock the last digit that you write is allowed to float, so `~> 1.2.0` is the same as `>= 1.2.0` + `< 1.3.0`, whereas `~> 1.2` is `>= 1.2.0` + `< 2.0`.

mwrock:
ok. thanks @danielsdeleo! The constraint to use here then does seem obvious. However we would need to change the code around the exception as mentioned above.

We use our chef-provisioning-vsphere driver to power our kitchen-vsphere driver. I have a forked chef-provisioning gem where I changed the dependency constraint so that winrm 1.3 gem can coexist. However it breaks in `winrm.rb`

mwrock:
ok. Just gave this another shot changing the constraint to `~>1.2` and changing the logic around error handling. There may very well be a more elegant way of handling the error based on winrm gem version but this manages to avoid rescuing `StandardError`.

Thanks again for teaching me stuff today!

jkeiser:
This looks good.  I think @tyler-ball is right and `~> 1.3` will result in a cleaner fix, but this OK to merge too.  I like your change to group the network errors.

lib/chef/provisioning/transport/winrm.rb
```diff
@@ -86,14 +86,14 @@ def available?
...
         true
-      rescue Timeout::Error, Errno::EHOSTUNREACH, Errno::ETIMEDOUT, Errno::ECONNREFUSED, Errno::ECONNRESET, ::WinRM::WinRMHTTPTransportError, ::WinRM::WinRMWebServiceError, ::WinRM::WinRMWSManFault
-        Chef::Log.debug("unavailable: network connection failed or broke: #{$!.inspect}")
-        disconnect
-        false
       rescue ::WinRM::WinRMAuthorizationError
         Chef::Log.debug("unavailable: winrm authentication error: #{$!.inspect} ")
         disconnect
         false
+      rescue network_errors
```
danielsdeleo: You can "splat" these so you don't have to make a container module below. For example:

```ruby
class A < StandardError; end
class B < StandardError; end
class C < StandardError; end
errors = [ A, B ]

begin
  raise A, "a"
rescue *errors => e
  puts e
end
# => a
begin
  raise B, "b"
rescue *errors => e
  puts e
end
# => b
begin
  raise C, "c"
rescue *errors => e
  puts e
end
# => does not rescue
```

mwrock:
Thanks @danielsdeleo. That is much more clear than the dynamic module.

I sort of glossed over @tyler-ball 's full comment yesterday and I agree with @jkeiser tha `~>1.3` is better as long as it does not disrupt the rest of the dk. It also makes breaking out the error group unnecesary and we can just rescue `WinRMError` and the other couple non-winrm errors which I think is clearer.

I can update the PR with this later today.

jkeiser:
It's worth pointing out that this loop is catching specific errors on
purpose ... if we switch to just catching a single error, there could be
permanent issues (like a missing credentials file locally) that we will
spin around waiting for. I'd rather not block this patch on it unless there
is a direct replacement for the now missing error, but we need to file an
issue (I am in transit and hoping I won't forget by the time I get to work.)

mwrock:
Apologies for not being more clear. I agree we should continue to rescue the auth error separately. I'm proposing changing the rescue block to:

```
       rescue ::WinRM::WinRMAuthorizationError
         Chef::Log.debug("unavailable: winrm authentication error: #{$!.inspect} ")
         disconnect
         false
      rescue Timeout::Error, Errno::EHOSTUNREACH, Errno::ETIMEDOUT, Errno::ECONNREFUSED, Errno::ECONNRESET, ::WinRM::WinRMError
        Chef::Log.debug("unavailable: network connection failed or broke: #{$!.inspect}")
        disconnect
        false
```

tyler-ball:
:+1: on @mwrock's proposed rescue block and pinning to `~> 1.3`

jkeiser:
@mwrock the concern is that replacing `, ::WinRM::WinRMHTTPTransportError, ::WinRM::WinRMWebServiceError, ::WinRM::WinRMWSManFault` with `::WinRM::WinRMError` is going to catch errors we didn't catch before from WinRM, and that some of those may be things we *don't* want to retry for.

However, I really really don't want you to stop on this patch, let's check in what you are planning and figure out what the right replacements are later.

mwrock:
I just updated the PR to reflect the block above. The WinRMError will currently catch these erors:

```
WinRMWSManFault
WinRMWMIError
WinRMHTTPTransportError
```

`WinRMAuthorizationError` also derrives from this but we should catch it specifically since it is rescued first. So I think that at least right now, `WinRMError` is at the granularity we want.

danielsdeleo:
Sounds like this patch now handles errors in the way we want it to?

jkeiser:
:+1:

tyler-ball:
:+1:

Issue #336, opened 2015-05-04
chef-dk 0.5.1 with chef-zero renders unusable chefzero://localhost:8889 URLs on nodes


causton81:
`chef-client -z -o some-app-recipe` will provision a node and render a /etc/chef/client.rb with a `chef_server_url chefzero://localhost:8889`

I worked around by adding `with_chef_server Chef::Config[:chef_server_url].sub('chefzero', 'http')`

PR #337, opened 2015-05-05
Updating for newly introduced socketless mode

tyler-ball:
fixes https://github.com/chef/chef-provisioning/issues/336

\cc @danielsdeleo @jkeiser 

I would rather not do a gsub, but this was a quick hack to get it working for me.  Is there some setting I can specify in `client.rb` on the remote machine that will make it ignore the `chefzero` part of the URL and just treat it like http?

### DIFF
--- a/lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb
+++ b/lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb
@@ -1,6 +1,7 @@
 require 'chef/provisioning/convergence_strategy'
 require 'pathname'
 require 'cheffish'
+require 'chef_zero/socketless_server_map'
 
 class Chef
 module Provisioning
@@ -187,6 +188,25 @@ module Provisioning
       end
 
       def client_rb_content(chef_server_url, node_name)
+        # Chef stores a 'port registry' of chef zero URLs.  If we set the remote host's
+        # chef_server_url to a `chefzero` url it will fail because it does not know
+        # about the workstation's chef zero server
+        uri = URI.parse(chef_server_url)
+        if uri.scheme == 'chefzero' && uri.host == 'localhost'
+          if !Chef::Config[:listen]
+            raise "The remote host is configured to access the local chefzero host, but " +
+              "the local chefzero host is not configured to listen.  Provide --listen or " +
+              "set `listen true` in the chef config."
+          end
+          # Once chef and chef-dk are using chef-zero which supports this, we can
+          # remove the else block and the if check
+          if ChefZero::SocketlessServerMap.respond_to?(:server_on_port)
+            chef_server_url = ChefZero::SocketlessServerMap.server_on_port(uri.port).url
+          else
+            chef_server_url = chef_server_url.gsub(/^chefzero/, 'http')
+          end
+        end
+
         ssl_verify_mode = convergence_options[:ssl_verify_mode]
         ssl_verify_mode ||= if chef_server_url.downcase.start_with?("https")
                               :verify_peer


danielsdeleo:
You should be able to check the chef config to see if local mode is enabled and then ask chef zero for the proper URL, no? If you can't do that now it shouldn't be that hard to update chef to allow that.  

I think you'd have to also consider the case where there's a non local mode server with a localhost address though.

danielsdeleo:
Also, you need to check if the user gave `--no-listen` or set `listen false` in the config, since in that case Chef Zero won't be listening on the socket at all.

tyler-ball:
First, a note: This only occurs when I'm using the ChefDK.  Just using chef 12.3.0 from github doesn't cause this issue, even when using it both locally and remotely.

Chef provisioning configures provisioned VMs to point to the same Chef server that the workstation is using.  If the workstation is using Chef Zero, we forward a port from the provisioned VM back to the workstation so it can connect to the workstation's Chef Zero server.

We need to update chef-provisioning to disable (or throw an error) if `--no-listen` is configured because then the provisioned machine won't be able to call back to the workstation's Chef Zero server.  I can do that as part of this PR.

If running `chef-client -z` the provisioned machine ends up with `chef_server 'chefzero://localhost:8889'` in its client.rb.  What I don't understand is why that is not working - it fails out with the error `ChefZero::ServerNotFound: No socketless chef-zero server on given port 8889
`.  The stacktrace on the provisioned machine is:

```
Generated at 2015-05-05 20:45:41 +0000
ChefZero::ServerNotFound: No socketless chef-zero server on given port 8889
/opt/chef/embedded/lib/ruby/gems/2.1.0/gems/chef-zero-4.2.1/lib/chef_zero/socketless_server_map.rb:78:in `request'
/opt/chef/embedded/lib/ruby/gems/2.1.0/gems/chef-zero-4.2.1/lib/chef_zero/socketless_server_map.rb:33:in `request'
/opt/chef/embedded/apps/chef/lib/chef/http/socketless_chef_zero_client.rb:153:in `request'
/opt/chef/embedded/apps/chef/lib/chef/http.rb:266:in `block in send_http_request'
/opt/chef/embedded/apps/chef/lib/chef/http.rb:298:in `block in retrying_http_errors'
/opt/chef/embedded/apps/chef/lib/chef/http.rb:296:in `loop'
/opt/chef/embedded/apps/chef/lib/chef/http.rb:296:in `retrying_http_errors'
/opt/chef/embedded/apps/chef/lib/chef/http.rb:260:in `send_http_request'
/opt/chef/embedded/apps/chef/lib/chef/http.rb:143:in `request'
/opt/chef/embedded/apps/chef/lib/chef/rest.rb:119:in `get'
/opt/chef/embedded/apps/chef/lib/chef/node.rb:499:in `load'
/opt/chef/embedded/apps/chef/lib/chef/node.rb:483:in `find_or_create'
/opt/chef/embedded/apps/chef/lib/chef/policy_builder/expand_node_object.rb:106:in `load_node'
/opt/chef/embedded/apps/chef/lib/chef/client.rb:239:in `load_node'
/opt/chef/embedded/apps/chef/lib/chef/client.rb:443:in `run'
/opt/chef/embedded/apps/chef/lib/chef/application.rb:271:in `block in fork_chef_client'
/opt/chef/embedded/apps/chef/lib/chef/application.rb:259:in `fork'
/opt/chef/embedded/apps/chef/lib/chef/application.rb:259:in `fork_chef_client'
/opt/chef/embedded/apps/chef/lib/chef/application.rb:225:in `block in run_chef_client'
/opt/chef/embedded/apps/chef/lib/chef/local_mode.rb:39:in `with_server_connectivity'
/opt/chef/embedded/apps/chef/lib/chef/application.rb:213:in `run_chef_client'
/opt/chef/embedded/apps/chef/lib/chef/application/client.rb:402:in `block in interval_run_chef_client'
/opt/chef/embedded/apps/chef/lib/chef/application/client.rb:392:in `loop'
/opt/chef/embedded/apps/chef/lib/chef/application/client.rb:392:in `interval_run_chef_client'
/opt/chef/embedded/apps/chef/lib/chef/application/client.rb:382:in `run_application'
/opt/chef/embedded/apps/chef/lib/chef/application.rb:60:in `run'
/opt/chef/embedded/apps/chef/bin/chef-client:26:in `<top (required)>'
/usr/bin/chef-client:51:in `load'
/usr/bin/chef-client:51:in `<main>'
```

The `--no-listen` setting does not apply to the provisioned machine because it is not running Chef Zero - only the workstation is.  Thats why I'm asking if there is a setting that will force chef-client to interpret `chefzero://localhost:8889` as still needing a socket.  In the provisioning case, it still needs to try and connect to that socket which is forwarded to the workstation.

tyler-ball:
I just tried running `chef-client -z --listen` on my workstation, and the remote VM still has the same issue.  The issue is not that my local Chef Zero instance isn't listening.

danielsdeleo:
@tyler-ball the behavior you see is what I would expect. Chef Zero has a registry of servers by port number, so you can have multiple chef zeros running at the same time, with different config. So when the provisioned machine gets the `chefzero://localhost:8889` URL and tries to make a request to it, there's going to not be a chef zero in the "port registry," therefore it errors out.

My comment about `--listen` vs. `--no-listen` is in addition to the patch you already made. The patch you proposed looks to me like it should work with the default settings, but the user could additionally specify `--no-listen` (currently not the default) in which case the provisioned system would fail when trying to contact the local machine's Chef Zero. In other words, it's just a case of invalid input from the user you need to detect.

tyler-ball:
@danielsdeleo Thanks for that clarification - I couldn't remember exactly how socketless mode works.  Because it stores a 'port registry' I think the gsub I proposed is okay, now that I added some comments around it.

I decided to move all the logic down into the `client_rb_content` method.  I could have kept the error checking logic in the initialization, but because this logic seems weird I decided to keep it all together.  What do you think?

danielsdeleo:
> In doing this I was thinking the fix I would rather do would be to have Chef attempt to access any chefzero:// urls (even if they don't yet exist in the registry) and automatically add it to the registry of the request is successful. How do you feel about that?

Lemme clarify. You're saying that something in Chef Client triggers an HTTP request, and the `chef_server_url` is 'chefzero://localhost:1234'. This request hits the socketless code, then Chef Zero says there's no such local mode server for port 1234, so then the request is retried at `http://localhost:1234` ?

tyler-ball:
@danielsdeleo Correct - thanks for parsing my terrible end-of-workday syntax.

danielsdeleo:
That feels a little too magic and special-case-y to me. At the time that Chef Provisioning runs, it can tell whether or not it's in local mode, because it has access to the provisioner-chef-client's config. You can also ask the local mode server what its URL is and generally do what's required to give the provisioned machine the exactly right Chef config, whereas on the provisioned client, we're basically just guessing. In the success case everything is fine, but I think it makes failure cases (granted, this might mainly apply to development, but that's important, too) a lot more confusing.

tyler-ball:
Updated to try and use the code from https://github.com/chef/chef-zero/pull/131 if its available, and only fall back to `gsub!` as a last resort

danielsdeleo:
:+1:

christinedraper:
Can you confirm whether this is a problem with knife -z bootstrap, as well as chef-provisioning? I get the same symptoms... client.rb contains chefzero://localhost:8889, and chef run fails with the 'no socketless chef-zero server".  I've figured out how to workaround it for chef-provisioning, but I can't figure out an equivalent for knife -z bootstrap.

danielsdeleo:
I don't believe `knife bootstrap` has ever had support for local mode. In order for that to work, `knife bootstrap` would need to create a reverse SSH tunnel from the bootstrap target back to the the workstation running knife (which is what Chef Provisioning does). I scanned the code and I did not find any support for this in the relevant places. But maybe there is some other aspect of your configuration that made this work previously that I'm not considering. In any case, that would be a chef-client issue, can you create a new issue there with the relevant details, especially anything special you did to set up port forwarding?

afiune:
Just for people out there that is using `chefdk` & `bundler` at the same time. :facepunch: (dunno why you would do that but this will help) You must update your gems in `bundler` since those are the ones that take precedence.

PR #339, opened 2015-05-08
Add options[:scp_temp_dir] to set a SCP destination other than /tmp

glennmatthews:
In one of our SSH provisioning scenarios, `/tmp` is not a valid location for `scp` to upload files to on the target machine. I've added a new option, `options[:scp_temp_dir]` which can be used to specify a different destination directory. This solves the problem for us.

````
with_machine_options :transport_options => {
...
  :options => {
    ...
    :scp_temp_dir => '/bootflash', # upload files to /bootflash instead of /tmp
  },
  ...
}
````

### DIFF
--- a/lib/chef/provisioning/transport/ssh.rb
+++ b/lib/chef/provisioning/transport/ssh.rb
@@ -27,6 +27,9 @@ module Provisioning
       #     nil (the default) means no gateway. If the username is omitted,
       #     then the default username is used instead (i.e. the user running
       #     chef, or the username configured in .ssh/config).
+      #   - :scp_temp_dir: a directory to use as the temporary location for
+      #     files that are copied to the host via SCP.
+      #     Only used if :prefix is set. Default is '/tmp' if unspecified.
       # - global_config: an options hash that looks suspiciously similar to
       #   Chef::Config, containing at least the key :log_level.
       #
@@ -38,6 +41,7 @@ module Provisioning
         @username = username
         @ssh_options = ssh_options
         @options = options
+        @scp_temp_dir = options.fetch(:scp_temp_dir, '/tmp')
         @config = global_config
       end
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by glennmatthews** [comment]
@mattray

File: lib/chef/provisioning/transport/ssh.rb [hunk 1]
```diff
@@ -107,7 +111,7 @@ def write_file(path, content)
         execute("mkdir -p #{File.dirname(path)}").error!
         if options[:prefix]
           # Make a tempfile on the other side, upload to that, and sudo mv / chown / etc.
-          remote_tempfile = "/tmp/#{File.basename(path)}.#{Random.rand(2**32)}"
+          remote_tempfile = "#{@scp_temp_dir}/#{File.basename(path)}.#{Random.rand(2**32)}"
```
tyler-ball: I would prefer `File.join(scp_temp_dir, "#{File.basename(path)}.#{Random.rand(2**32)})`

lib/chef/provisioning/transport/ssh.rb
```diff
@@ -38,6 +41,7 @@ def initialize(host, username, ssh_options, options, global_config)
         @username = username
         @ssh_options = ssh_options
         @options = options
+        @scp_temp_dir = options.fetch(:scp_temp_dir, '/tmp')
```
tyler-ball: You should add an `attr_reader` for this on line 53

lib/chef/provisioning/transport/ssh.rb
tyler-ball: Actually, because we reference `remote_tempfile` multiple places and they all have the same implementation, it should just be extracted to a method.

tyler-ball:
Hey @glennmatthews thanks for this PR.  We think the idea is great, but I just had a few minor code updates.  Once those are in this will be good to merge!

glennmatthews:
Added attr_reader for scp_temp_dir, added remote_tempfile() method. Thanks for the review!

tyler-ball:
@glennmatthews I made some minor updates while merging, but thanks for this initial PR!

glennmatthews:
Thanks for the improvements. Much appreciated!

lib/chef/provisioning/transport/ssh.rb
poliva83: @tyler-ball @glennmatthews I think there may be issue with this line.

My co-worker @andrewjamesbrown was hitting error which latest chefdk 0.7.0 that I was unable to produce on 0.6.2. It seems when you have have specified machine_options attribute in machine resource block then @options becomes a Cheffish::MergedConfig object. 

https://github.com/chef/chef-provisioning/blob/master/lib/chef/resource/machine.rb#L101

This object has not implemented the fetch method so we're seeing an exception.

PR #345, opened 2015-05-22
with_driver must be specified

jtimberman:
If with_driver is not specified, trying to use this recipe with a machine resource will result in a RuntimeError:

```
Driver not specified for machine mario
```

Where "mario" is from the "simple" recipe example in this repository.


scotthain:
:+1:

PR #346, opened 2015-05-24
Fix chef_group

obazoud:


tyler-ball:
Thanks for the catch @obazoud!

PR #364, opened 2015-06-16
Fix #358: Arity mismatch

### LINKED ISSUE #358: Arity mismatch between MachineImage's :destroy and Driver#destroy_image
arity-mismatch

Since this is a published API and third-party drivers might already be calling `Driver#destroy_image` with 3 arguments, the solution is to change

```ruby
def destroy_image(action_handler, image_spec, image_options, machine_options)
```

to

```ruby
def destroy_image(action_handler, image_spec, image_options, machine_options={})
```

(It's a separate question of whether `machine_options` makes sense here.)

In the meantime, drivers not implementing  `#destroy_image` can work around this by overriding the superclass method:

```ruby
def destroy_image(action_handler, image_spec, image_options, machine_options={})
    super(action_handler, image_spec, image_options, machine_options)
end
```

Which replaces the `ArgumentError` with a `RuntimeError` from the superclass which will give a correct and useful error message.

randomcamel:


randomcamel:
Fixes #358.

fnichol:
LGTM! :+1:

randomcamel:
:shipit:

PR #379, opened 2015-07-01
Deleting acceptance tests since they are super stale, fixes #363

### LINKED ISSUE #363: remove or update test/acceptance project
This was something we had experimented with last year, and the work was left as it is due to priority work.  This is now out of date, and the momentum to improve it as we built it was lost.  The spec work in the aws driver was the direction we are planning on going at the time.  My vote would be get rid of the clutter.

tyler-ball:
\cc @patrick-wright @jkeiser @metadave @randomcamel 

Fixes #363


patrick-wright:
DO EET!

randomcamel:
:+1: Deleted code is tested code.

PR #382, opened 2015-07-02
Create client.rb only if missing. Fixes #318

### LINKED ISSUE #318: Don't continually manage `client.rb`.
Chef provisioning should lay down `client.rb` only once. The issue we're running in to is Provisioning comes along and creates `client.rb` then does the first converge which runs our recipes to configure chef-client further, including add other options to `client.rb`. Then we run into a scenario where Provisioning and the chef client daemon fight over the configuration continually changing it back and forth between what Provisioning and the daemon want. I think Chef Provisioning should only manage this file once and then leave it to the node's recipes to determine.

dblessing:
This seems like a straightforward fix for #318. Previously, the `client.rb` file checksum was compared on each `chef-provisioning` run and if it didn't match it was updated. This caused a fight if the node was also running `chef-client` daemon - the daemon might change the client.rb file contents which makes `chef-provisioning` unhappy. I really believe that `chef-provisioning` should only manage the `client.rb` on the first run.

### DIFF
--- a/lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb
+++ b/lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb
@@ -27,7 +27,9 @@ module Provisioning
 
         # Create client.rb and client.pem on machine
         content = client_rb_content(chef_server_url, machine.node['name'])
-        machine.write_file(action_handler, convergence_options[:client_rb_path], content, :ensure_dir => true)
+        unless machine.file_exists?(convergence_options[:client_rb_path])
+          machine.write_file(action_handler, convergence_options[:client_rb_path], content, :ensure_dir => true)
+        end
       end
 
       def converge(action_handler, machine)


jkeiser:
What happens if the user changes the Chef config options, or the key, or the server? I think we could be more discriminating about it, but the contract of the resource is that it will repair anything that contradicts what is written the resource, which includes client.rb.

One option is to parse the config file and see if the options match what need to be there. Another The second best thing would be to have an option to stay hands off the config file, and flip that option on by default if the Chef client recipe is in the node's run list.  I just don't think we can turn it off unconditionally.

dblessing:
I like the option to toggle client.rb management (except if it isn't there). 

This is something I'd love to help solve very quickly. Now that we have many nodes the provisioner chef runs are converging every node every time because of this behavior. It doesn't scale very well. Thanks for your feedback.

jkeiser:
I totally understand! As unhappy as I generally am about toggles, the perfect solution seems fraught with difficulty, and the toggle should get the job done much more quickly.  (If this toggle detects the presence of the Chef client recipe in the run list, it's about as automatic as I can think, too, and that's my usual hatred of toggles--making the user do something your software could have figured out for itself.)

tyler-ball:
What about adding a new action, `converge_if_missing` and making that the default instead of `converge`?  Changing the default would require a major version change, but we could certainly add a new action inside a minor release.

poliva83:
Quick and dirty way but can't you change `/etc/chef/client.rb` using files attribute in machine resource to content that makes both chef-provisioning and chef-client daemon happy?

```
machine vm_name do
  chef_environment "_default"
  recipe "openssh::default"
  tag ""
  files ({
    "/etc/chef/client.rb" => { :content => <<END_CLIENT_RB
chef_server_url "#{chef_server_url}"
node_name "#{vm_name}"
validation_client_name "chef-validator"
client_key "/etc/chef/client.pem"
ssl_verify_mode :verify_none

Dir.glob(File.join("/etc/chef", "client.d", "*.rb")).each do |conf|
Chef::Config.from_file(conf)
end

END_CLIENT_RB
    }
  })
  action :converge
end
```

Looking at `client_rb_content()` method in `PrecreateChefObjects` class it seems like you might be able to somewhat control `/etc/chef/client.rb` content through passing machine option :chef_config. Maybe @jkeiser can clarify if this is true or I am reading code wrong.  

https://github.com/chef/chef-provisioning/blob/master/lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb#L230

But certain lines in file it seems you can't control currently (`chef_server_url`, `node_name`, etc)?

dblessing:
We tried the latter method and couldn't quite get it. The first method you mention may work, though. If specifying client.rb using the file attribute will provisioning fight itself because the md5 won't match what it expects?

poliva83:
Your right that chef-provisioning does still fight with itself even using first method. Below are logs of my 2nd chef-provisioning convergence in debug mode.

So it would be possible for the daemon process to fire at the moment and have a chef-client run fail because of bad conifg but won't it also fail if it attempts to run while chef-provisioning is converging? Since only 1 chef-client process allowed at any time. I think you may need to put logic in your chef-provisioning script to disable chef-client daemon then let the cookbook re enable it during convergence caused by chef-provisioning.

```
[2015-07-25T15:59:29+00:00] DEBUG: Forwarding local server localhost:7799 to local@10.236.209.143
[2015-07-25T15:59:29+00:00] INFO: Port forwarded: local URL http://localhost:7799 is available to 10.236.209.143 as http://localhost:7799 for the duration of this SSH connection.
[2015-07-25T15:59:29+00:00] INFO: Executing sudo ls -d /etc/chef/client.rb on local@10.236.209.143
/etc/chef/client.rb
[2015-07-25T15:59:29+00:00] INFO: Completed ls -d /etc/chef/client.rb on local@10.236.209.143: exit status 0
[2015-07-25T15:59:29+00:00] INFO: Executing sudo md5sum -b /etc/chef/client.rb on local@10.236.209.143
8e0aa9cabba8cd0384d39e62e8ec2c64 */etc/chef/client.rb
[2015-07-25T15:59:29+00:00] INFO: Completed md5sum -b /etc/chef/client.rb on local@10.236.209.143: exit status 0
[2015-07-25T15:59:29+00:00] INFO: Executing sudo ls -d /etc/chef on local@10.236.209.143
/etc/chef
[2015-07-25T15:59:29+00:00] INFO: Completed ls -d /etc/chef on local@10.236.209.143: exit status 0
[2015-07-25T15:59:29+00:00] INFO: Executing sudo mkdir -p /etc/chef on local@10.236.209.143
[2015-07-25T15:59:29+00:00] INFO: Completed mkdir -p /etc/chef on local@10.236.209.143: exit status 0
[2015-07-25T15:59:29+00:00] DEBUG: Writing 142 bytes to /tmp/client.rb.473181203 on local@10.236.209.143
[2015-07-25T15:59:29+00:00] INFO: Executing sudo mv /tmp/client.rb.473181203 /etc/chef/client.rb on local@10.236.209.143
[2015-07-25T15:59:29+00:00] INFO: Completed mv /tmp/client.rb.473181203 /etc/chef/client.rb on local@10.236.209.143: exit status 0

    - write file /etc/chef/client.rb on poliva-bescloud-admin-node
```

tyler-ball:
I agree with @dblessing that `chef-provisioning` shouldn't be trying to manage the `client.rb` (or any other files on the system, including pem files or other config files) if `chef-client` is managing those files.

I think it sounds nice to inspect the node and only run the `precreate_chef_objects` logic if the node isn't currently being managed by chef.  But its hard to actually determine if the node _is being managed by chef_.  We could look for the presence of a `client.rb` file, but what if users are storing the file in a custom location for security reasons and specifying it with the `-c` flag when running `chef-client`?

I think a better way would be to inspect the chef server and determine if there is already a `client` object for the node in question.  What do you think of that?

I still think we would want either a flag or multiple actions.  I can see users wanting to force a 're-provision' by forcing the `precreate_chef_logic` and re-registering the client with the server.  I think they will want this regardless of how smart we make the detection logic.

tyler-ball:
Pinging this for feedback on the following 2 issues:

> I think a better way would be to inspect the chef server and determine if there is already a client object for the node in question. What do you think of that?

I also think that the `converge_if_missing` action is the best way to make it configurable.  That way users can still use the `converge` action if they want to force convergence.

There is also the `converge` attribute that we could modify.  We could make it accept a symbol instead of true/false.  Then users could provide `:always`, `:if_missing` or `:never`.  Then `true`/`false` could be cast to `:always`/`:never`.

tyler-ball:
I'm going to close this PR since I think there are better ways we want to solve this issue.  I've moved some comments into the original issue so we can tackle that when we are ready.  Thanks for the stab at it though @dblessing !

PR #389, opened 2015-07-14
use mixlib-install to install from omnitruck

### LINKED ISSUE #300: chef_version convergence option is ignored for windows machines
Unlike the Linux [install.sh](https://github.com/chef/chef-provisioning/blob/master/lib/chef/provisioning/convergence_strategy/install_sh.rb) convergence strategy, the Windows [install.msi](https://github.com/chef/chef-provisioning/blob/master/lib/chef/provisioning/convergence_strategy/install_msi.rb) strategy does not take into account the `chef_version` in `convergence_options`, so Windows nodes always have the latest Chef client installed. Of course, this can be problematic when the latest version introduces bugs as is sometimes the case.

### LINKED ISSUE #327: Windows hosts are bootstrapped with the wrong url for chef-client
For Windows hosts, it appears we use the url `'https://www.chef.io/chef/install.msi'`. This retrieves the latest version of chef-client for Windows, but the correct source is actually omnitruck. This is what `knife-windows` does, see https://github.com/chef/knife-windows/blob/0.8.4/lib/chef/knife/bootstrap/windows-chef-client-msi.erb#L110. Alternatively, `chef-provisioning` could simply use `knife-windows` to perform the bootstrap since it also handles authentication / SSL configuration.

Without this fix, provisioning depends on a url that may go away and also does not allow for the retrieval of a specific version. Provisioning currently doesn't honor the convention in knife bootstrap that nodes be bootstrapped with the latest minor version of the version of of Chef running knife, for instance.

thommay:
closes #327, closes #300, closes #38

chef/mixlib-install is extracted from test-kitchen. It supports installing nightlies and prereleases, supports windows properly and doesn't use bash.

### DIFF
--- a/chef-provisioning.gemspec
+++ b/chef-provisioning.gemspec
@@ -19,6 +19,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'inifile', '~> 2.0'
   s.add_dependency 'cheffish', '~> 1.1'
   s.add_dependency 'winrm', '~> 1.3'
+  s.add_dependency "mixlib-install",  "~> 0.4"
 
   s.add_development_dependency 'rspec'
   s.add_development_dependency 'rake'
--- a/lib/chef/provisioning/convergence_strategy/install_msi.rb
+++ b/lib/chef/provisioning/convergence_strategy/install_msi.rb
@@ -1,4 +1,5 @@
 require 'chef/provisioning/convergence_strategy/precreate_chef_objects'
+require 'mixlib/install'
 require 'pathname'
 
 class Chef
@@ -9,8 +10,6 @@ module Provisioning
         super
         @chef_version ||= convergence_options[:chef_version]
         @prerelease ||= convergence_options[:prerelease]
-        @install_msi_url = convergence_options[:install_msi_url] || 'https://www.chef.io/chef/install.msi'
-        @install_msi_path = convergence_options[:install_msi_path] || "$env:TEMP\\#{File.basename(@install_msi_url)}"
         @chef_client_timeout = convergence_options.has_key?(:chef_client_timeout) ? convergence_options[:chef_client_timeout] : 120*60 # Default: 2 hours
       end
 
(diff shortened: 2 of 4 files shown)


thommay:
cc @lamont-granquist @fnichol

chef-provisioning.gemspec
```diff
@@ -19,6 +19,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'inifile', '~> 2.0'
   s.add_dependency 'cheffish', '~> 1.1'
   s.add_dependency 'winrm', '~> 1.3'
+  s.add_dependency 'mixlib-install'
```
tyler-ball: This should have a major and minor version constraint in here

lib/chef/provisioning/convergence_strategy/install_msi.rb
```diff
@@ -9,8 +10,6 @@ def initialize(convergence_options, config)
         super
         @chef_version ||= convergence_options[:chef_version]
         @prerelease ||= convergence_options[:prerelease]
-        @install_msi_url = convergence_options[:install_msi_url] || 'https://www.chef.io/chef/install.msi'
```
tyler-ball: If we just stop using these options, we will have to move this change to a major release version.  Otherwise we need to support them for backwards compatibility.  Does `Mixlib::Install` not have analogous configs?

lib/chef/provisioning/convergence_strategy/install_sh.rb
```diff
@@ -29,58 +30,19 @@ def initialize(convergence_options, config)
...
-            script    = proxy.start(chef_uri.host, :use_ssl => proxy_uri.scheme == 'https') do |http|
-              http.request(req)
-            end
-            script.body
-          end
-        end
-        machine.write_file(action_handler, install_sh_path, @@install_sh_cache[install_sh_url], :ensure_dir => true)
-        # TODO handle bad version case better
-        machine.execute(action_handler, install_sh_command_line)
+        install_command = Mixlib::Install.new(chef_version, false, opts).install
```
tyler-ball: `install` seems like a misleading method name if all it does is return a string that then needs to be executed to perform the actual install.

randomcamel:
Can there be a test for this?

tyler-ball:
Hey @thommay - we're doing bug triage right now.  Do you need any help with the comments we made for this?  I'm in full support of the idea (leveraging shared install logic between Knife and chef-provisioning)!

chef-provisioning.gemspec
thommay: Not sure there's a useful version constraint to be had right now - I can lock it to `<= 1.0` but that doesn't actually provide any useful dependency data.

lib/chef/provisioning/convergence_strategy/install_msi.rb
thommay: Yeah, we can support the option still, but the behaviour will change a little - rather than always installing the latest version, windows will now install the version the user tells us to if they don't give a specific msi install url. 
If they do, we can just grab that file as before.

thommay:
@randomcamel there don't seem to be tests for any of chef-provisioning, so I didn't want to go down that rabbit hole in this PR. Mixlib::Install itself has a fair number of tests and I'm happy to add more.

thommay:
@tyler-ball sorry, this disappeared into email somewhere. I've just pushed chef/mixlib-install#2 which restores the ability to use a specific msi install url, which i'll merge and release momentarily.

chef-provisioning.gemspec
```diff
@@ -19,6 +19,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'inifile', '~> 2.0'
   s.add_dependency 'cheffish', '~> 1.1'
   s.add_dependency 'winrm', '~> 1.3'
+  s.add_dependency "mixlib-install",  ">= 0.4", "< 1.0"
```
tyler-ball: You could use `~> 0.4` and it will accomplish the same thing as what you have here

lib/chef/provisioning/convergence_strategy/install_msi.rb
tyler-ball: Sounds like the right way to handle this to me!

thommay:
@tyler-ball ah, that wasn't clear to me from the docs. Thanks, updated the dependency.

tyler-ball:
The code LGTM.  Have you done a manual test of both install methods?  If not I can help you do that using chef-provisioning-aws.  Once a smoke test is done I'll merge this.

thommay:
Yeah, tested with windows with both metadata and an msi url; ubuntu and omnios - https://gist.github.com/thommay/153681df62bc012da685 is the code, for the record.

tyler-ball:
@thommay if you rebase off master there is now a .travis.yml which configures this to not use ruby 1.9 - should fix the travis error and then we can merge

thommay:
@tyler-ball done and done, thanks.

PR #404, opened 2015-08-04
Zz/policy support

double-z:
This pull request adds policyfile support - beyond passing it in in the chef_config convergence options - to  the machine resource/provider vi the use of `policy_name` and `policy_group` machine resource attributes, which are then handled within the convergence_strategy `precreate_chef_objects.rb` during the client_rb_content method.

Not sure if this is something we want to introduce at this point, with policyfiles still being experimental, so this is at minimum a way to begin the discussion. 

This strikes me as how to go about it, for now at least, but thoughts on other approaches in the comments here would be great.

The usage is as follows:

```
machine "sshone" do
  action :converge
  machine_options 'transport_options' => {
    'ip_address' => '192.168.33.22',
    'username' => 'vagrant',
    'ssh_options' => {
      'password' => 'vagrant'
    }
  }
  policy_name 'machine_test'
  policy_group 'policymachine'
end
```

It requires native policyfile support so 12.1+ server or hosted is required, i.e the `with_local_chef_server` option isn't supported at this time, because  `policy_document_native_api` and `policy_group` along with the `-z` option for chef-client are all not supported in zero.

Some features include:
- sanity check that checks if running provisioning in local mode.
- sanity check to ensure `convergence_strategy[:chef_config]` does not contain any conflicting policy settings
- sanity check to ensure both `policy_group` and `policy_name` have been passed.

All of the above have been tested against all possible relative situations and behave as expected. Not using `policy_name` and `policy_group` has no affect on overall behavior and are silently ignored in the same way as other undefined convergence_options... It is only injected into the convergence_strategy if it is defined and properly defined.


chef-supermarket:

* @double-z

chef-supermarket:

lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb
```diff
@@ -193,6 +193,12 @@ def client_rb_content(chef_server_url, node_name)
         # about the workstation's chef zero server
         uri = URI.parse(chef_server_url)
         if uri.scheme == 'chefzero' && uri.host == 'localhost'
+          if convergence_options[:policy_name] || convergence_options[:policy_group]
+            raise "You are running Chef Zero and have policyfiles enabled " +
+              "however Chef Zero does not currently support the use " +
+              "of policyfile nativemode which is required for this option. " +
+              "To use this feature you must use chef-server 12.1 plus or hosted chef."
```
marcparadise: Minor nitpick, "Chef Server 12.1 or later, or Hosted Chef.

marcparadise: Would it make sense to provide a link to get chef server as well, https://downloads.chef.io/chef-server/?

double-z: good point

lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb
```diff
@@ -226,10 +234,43 @@ def client_rb_content(chef_server_url, node_name)
...
+          EOM
+        end
+
         content.gsub!(/^\s+/, "")
         content << convergence_options[:chef_config] if convergence_options[:chef_config]
         content
       end
+
+      def validate_policy_if_enabled
+        raise "Policy_name was given but no Policy_group. Both are required" if (convergence_options[:policy_name] &&
```
marcparadise: `'policy_name'` instead of Policy_name, since that's the name of the actual attribute/option. Same for `'policy_group`'

lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb
double-z: yeah, good thinking on the link. easify usage.

lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb
double-z: and another good point, consistency to avoid confusion.

lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb
```diff
@@ -226,10 +238,43 @@ def client_rb_content(chef_server_url, node_name)
...
           EOM
         end
+        if convergence_options[:policy_name]
+          content << <<-EOM
+            # Policyfile Settings:
+            use_policyfile true
+            policy_document_native_api true
+            policy_group "#{convergence_options[:policy_group]}"
+            policy_name "#{convergence_options[:policy_name]}"
+          EOM
```
jtimberman: You'll want to either remove the leading whitespace from these, or do something like this:

```ruby
content <<-EOM.gsub(/^\s+/, '')
```

Because the whitespace will be inserted in the rendered configuration.

tyler-ball: Agreed!

tyler-ball:
Hey @double-z - overall we think this PR is a good idea.  It adds new features that customers can choose to use.  But we do have some concerns:

1. We would also like to get tests started for this!  At the very least we can get a unit test added for the `client_rb_content` method.  You don't need to get full test coverage but it would be nice to see that started.
1. There is now a dummy driver that we can use to write integration tests.  Ping @randomcamel to help you get started on this.
1.  There also needs to be documentation.  Can you update the [configuration.md](https://github.com/chef/chef-provisioning/blob/master/docs/configuration.md#private-keys) document at the same level as the `Private Keys` and document this new feature?  You have a great writeup about its features in this PR description so that would be a good start.

lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb
double-z: correct me if i'm wrong here, but i think that is already handled globally on line 251 in my branch, 229 in master:

From my PR branch:

https://github.com/double-z/chef-provisioning/blob/zz/policy_support/lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb#L229-L253

From master:

https://github.com/chef/chef-provisioning/blob/master/lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb#L217-L231

if i need more than that lemme know... pretty sure that handles it already, but will test while doing below to be sure.

double-z:
got it @tyler-ball will have that asap... also wanna add a check to make sure we are on a policyfile supported non-zero chef-server ala the chef-zero check already included. will add that in as well unless i hear otherwise.

tyler-ball:
Hey @double-z - I see this hasn't been updated in a while.  Do you need some help continuing it?

lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb
tyler-ball: @double-z You are correct!

jtimberman:
This also needs to account node object changes from [this chef pull request](https://github.com/chef/chef/pull/3913)

danielsdeleo:
Yup, the reason this wasn't done before is because we wanted to have "native" support for policyfile attributes on the node object itself. The latest Chef Server nightlies support this (release version will be 12.3), and I am about to merge the pull request (see Joshua's comment) for Chef Client and knife to support it.

On the one hand, the implementation here would be nice, since it's compatible with earlier chef versions (i.e., the ones that are actually available now), but on the other hand maybe it would be weird if it eventually did both (set policy_name/policy_group in both the config and on the node)? I guess it could switch from the config file way to the node object way if it knows what version of chef you will have. Thoughts?

jtimberman:
Rather than checking the version of Chef, why not ask `Chef::Node` if it knows about `policy_name` and `policy_group`?

```ruby
Chef::Node.instance_methods.include?(:policy_name) && Chef::Node.instance_methods.include?(:policy_group)
```

danielsdeleo:
Yeah, or `respond_to?` would work as well, I think. Either way.

double-z:
hey @tyler-ball was actually putting together the spec tests and what not for this which can be seen here 

https://github.com/chef/chef-provisioning/blob/zz/policyfile_machine_resource_attributes/spec/chef/provisioning/convergence_strategy/precreate_chef_objects_spec.rb

was also testing out `deployment_group` in there too, but based on that testing and the above convo am pulling that out unless thoughts are i shouldn't. 

Need to look over the changes @jtimberman referenced and the suggestions he and @danielsdeleo have above. 

Was about to submit the final pull with tests included but in light of the above will work on the new additions this weekend and have that ready for monday.

tyler-ball:
Those tests LGTM.  Because you're making that test file from scratch you don't need to worry about getting 100% coverage of the whole class.  You just need to test your code changes.

lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb
```diff
@@ -226,10 +238,43 @@ def client_rb_content(chef_server_url, node_name)
...
         content << convergence_options[:chef_config] if convergence_options[:chef_config]
         content
       end
+
+      def validate_policy_if_enabled
+        raise "policy_name was given but no policy_group. Both are required" if (convergence_options[:policy_name] &&
+                                                                                 !convergence_options[:policy_group])
+        raise "policy_group was given but no policy_name. Both are required" if (convergence_options[:policy_group] &&
+                                                                                 !convergence_options[:policy_name])
+        if convergence_options[:policy_name] && convergence_options[:chef_config]
```
tyler-ball: This `if` check isn't necessary since you will have already raised an exception if this isn't true

lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb
```diff
@@ -226,10 +238,43 @@ def client_rb_content(chef_server_url, node_name)
...
+          entry_in_chef_config = []
+          ["use_policyfile", "policy_document_native_api", "policy_group", "policy_name"].each do |opts|
+            entry_in_chef_config << opts if convergence_options[:chef_config].include?(opts)
+          end
+
+          if !entry_in_chef_config.empty?
```
tyler-ball: I don't know if we need this logic here asserting that users didn't supply those configs in `:chef_config`.  What if they want to do something super weird like have `use_policyfile` reference a node attribute to determine if policyfiles are used?  I think we should keep our validation simple for now and add more validation as we se a need for it.

lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb
double-z: good point.

jtimberman:
I think we also need to support `named_run_lists`, which is probably best done by passing `-n` to the chef-client command in the convergence strategy code.

josb-vmware:
Any updates? Having better integration would make policies easier to use.

PR #408, opened 2015-08-09
Add a generic rspec module, and use and test it via the generator script.

randomcamel:


bin/generate_driver
```diff
@@ -139,14 +165,36 @@ end
...
-  # include #{camel_name}Config   # uncomment to get `chef_config` or to mix in code.
-
-  when_the_chef_12_server "exists", server_scope: :context, port: 8900..9000 do
-    with_#{snake_name} "integration tests" do
-      context "machine resource" do
-        it "doesn't run any tests" do
-        end
+  include_context "run with driver", :driver_string => "#{snake_name}"
+
+  with_chef_server do
```
tyler-ball: Does this need to be the `when_the_chef_12_server` like we have in all the AWS tests?  I like `with_chef_server` much better but I am curious about what the switch means

tyler-ball:
This is awesome.  I have 1 question but I am :+1: on this and you can merge it

bin/generate_driver
randomcamel: Nope, `with_chef_server` actually calls `when_the_chef_12_server`, but

1. with a clearer name, and
1. with default values.

All the `chef-provisioning-aws` invocations look like

    when_the_chef_12_server "exists", organization: 'foo', server_scope: :context

and this encapsulates (overridable) defaults

    organization: "spec_tests", server_scope: :context, port: 8900..9000

PR #409, opened 2015-08-11
Upload the installation script and run it from the host.

rancohen1:
the WinRM library base 64 encodes the powershell script and runs it inside a cmd using this one liner : "powershell -encodedCommand ...".
It has gotten too long and passes the 8191 character limit! until they come up with a better idea we can upload the script and then run it, or somehow run the script in a live session.

PR #410, opened 2015-08-12
Upload the PS1 script and run directly

thommay:
* Introduce the ability to run a cmd directly in winrm, rather than
  through powershell
* Arrange for the output of the install script to be streamed

Fixes #409, Fixes chef/chef-provisioning-aws#284

cc @rancohen1 @tyler-ball (This supercedes #409 and should be more robust)

### DIFF
--- a/lib/chef/provisioning/transport/winrm.rb
+++ b/lib/chef/provisioning/transport/winrm.rb
@@ -36,8 +36,11 @@ module Provisioning
       def execute(command, execute_options = {})
         output = with_execute_timeout(execute_options) do
           session.set_timeout(execute_timeout(execute_options))
-          session.run_powershell_script(command) do |stdout, stderr|
-            stream_chunk(execute_options, stdout, stderr)
+          block = Proc.new { |stdout, stderr| stream_chunk(execute_options, stdout, stderr) }
+          if execute_options[:raw]
+            session.run_cmd(command, &block)
+          else
+            session.run_powershell_script(command, &block)
           end
         end
         WinRMResult.new(command, execute_options, config, output)
--- a/lib/chef/provisioning/convergence_strategy/install_msi.rb
+++ b/lib/chef/provisioning/convergence_strategy/install_msi.rb
@@ -20,10 +20,11 @@ module Provisioning
 
       def setup_convergence(action_handler, machine)
         if !convergence_options.has_key?(:client_rb_path) || !convergence_options.has_key?(:client_pem_path)
-          system_drive = machine.execute_always('$env:SystemDrive').stdout.strip
+          system_drive = machine.system_drive
           @convergence_options = Cheffish::MergedConfig.new(convergence_options, {
             :client_rb_path => "#{system_drive}\\chef\\client.rb",
-            :client_pem_path => "#{system_drive}\\chef\\client.pem"
+            :client_pem_path => "#{system_drive}\\chef\\client.pem",
+            :install_script_path => "#{system_drive}\\chef\\\install.ps1"
           })
         end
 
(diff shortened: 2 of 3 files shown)


lib/chef/provisioning/transport/winrm.rb
```diff
@@ -36,8 +36,11 @@ def initialize(endpoint, type, options, global_config)
       def execute(command, execute_options = {})
         output = with_execute_timeout(execute_options) do
           session.set_timeout(execute_timeout(execute_options))
-          session.run_powershell_script(command) do |stdout, stderr|
-            stream_chunk(execute_options, stdout, stderr)
+          block = Proc.new { |stdout, stderr| stream_chunk(execute_options, stdout, stderr) }
+          if execute_options[:raw]
```
tyler-ball: Do you think there will be a 3rd way to run Windows commands in the future?  I've been converting a lot of Boolean flags to Symbols recently (to support more than 2 binary options) and if that will happen with this I would rather use a symbol flag now.

tyler-ball:
This LGTM.  I would love to get feedback from @smurawski and @fnichol - I know they had to do a bunch of workarounds in Test Kitchen to support transfer sizes (among other size limits)

lib/chef/provisioning/transport/winrm.rb
thommay: these are really the only two modes that WinRMI supports, so I'm not sure that there'll be more. But I can totally make it a symbol.

thommay:
So my suspicion is that it's the expansion of `$env:TEMP` that pushes us over the edge in terms of character limit, which is something that's hard to control. Uploading the file and then executing it seems - to me at least - a more robust mechanism, but I know nothing of the history of winrm...

ajeba99:
I added these changes to my environment and they test fine.  The only strangeness I saw was ...
````
    - write file C:\chef\install.ps1 on test-windows-server
    - write file C:\chef\install.ps1 on test-windows-server
````
It ran successfully though.

ajeba99:
@thommay So, after further testing.  The above changes work fine on a default 2012r2 AWS instance.

However they do not work on a default 2008r2 AWS instance with the following error ...
````
    - write file C:\chef\install.ps1 on Web2
    - write file C:\chef\install.ps1 on Web2
    [Web2] -----> Installing Chef Omnibus (always install latest version)

                  Downloading package from https://opscode-omnibus-packages.s3.amazonaws.com/windows/2008r2/x86_64/chef-
client-12.4.1-1.msi

                  Download complete.

                  Successfully verified C:\Users\ADMINI~1\AppData\Local\Temp\chef-latest.msi

                  Installing Chef Omnibus package C:\Users\ADMINI~1\AppData\Local\Temp\chef-latest.msi

    [Web2] msiexec was not successful. Received exit code
           At C:\chef\install.ps1:55 char:28
           +   if ($p.ExitCode -ne 0) { throw "msiexec was not successful. Received exit
           code ...
           +
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
               + CategoryInfo          : OperationStopped: (msiexec was not...ived exit c
              ode :String) [], RuntimeException
               + FullyQualifiedErrorId : msiexec was not successful. Received exit code


    ================================================================================
    Error executing action `converge` on resource 'machine[Web2]'
    ================================================================================

    RuntimeError
    ------------
    Error: command 'powershell.exe -ExecutionPolicy Unrestricted -NoProfile "& ""C:\chef\install.ps1"""' exited with cod
e 1.
````
After I upgraded the 2008r2 instance to Powershell 4.0 (manually) it worked fine.
Is there a way to make the above fix more Powershell 3.0 friendly?

lib/chef/provisioning/transport/winrm.rb
tyler-ball: I am ambivalent - if you don't think its likely then I would leave it alone.

smurawski:
I haven't had a chance to look at the core issue but if it doesn't work on 2008r2  with ps3 but does work with ps4 on the same os, then you are likely hitting the bug in winrm's configuration that I blogged about here -  http://stevenmurawski.com/powershell/2015/5/powershell-3-winrm-maxmemorypershellmb-is-wrong

There is a hotfix, but you have to request it. Better to just go to wmf4.

Now if it is stock 2008r2 image, the winrm settings likely need to be bumped up (see knife-windows).

thommay:
From talking to @smurawski more it sounds like the real fix here is to use a scheduled task to run the installer, which i'll look at next week

tyler-ball:
@thommay I went ahead and merged this since it will help with 2012, and we can address the scheduled task in another issue

PR #414, opened 2015-08-13
Adding ignore_converge_failure option, fixes #393

### LINKED ISSUE #393: specifying an audit-mode in a machine's run_list fails the provisioning run.
When I have an audit-mode recipe as part of my run_list in a provisioning recipe, the provisioning run stack traces with a generic

The audit-mode recipe is designed to intentionally fail audits, but as of chef-client 12.4.1, audit failures are not supposed to break a chef-client run; merely report that there are audit items that need remediation.

tyler-ball:
So a provisioned node that fails convergence does not kill the provisioning workstation run_list

\cc @jkeiser @ricardolupo @galenemery 

TODO: when this is merged update https://github.com/chef/chef-provisioning-aws#machine-options to point to this README


lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb
```diff
@@ -2,11 +2,20 @@
...
+require_relative 'ignore_convergence_failure'
 
 class Chef
 module Provisioning
   class ConvergenceStrategy
     class PrecreateChefObjects < ConvergenceStrategy
+
+      def initialize(convergence_options, config)
+        super
+        if convergence_options[:ignore_failure]
```
randomcamel: I think you're working on elaborating this, but it seems like we want to be able to ignore audit failures, non-audit failures, or both.

tyler-ball: `ignore_failure` can now accept a Fixnum, a Range, a custom error, or some array with a combination of those.  Or it can accept `true` to ignore all RuntimeErrors.  The documentation has been updated to show this.

lib/chef/provisioning/convergence_strategy/ignore_convergence_failure.rb
```diff
@@ -0,0 +1,53 @@
...
+          # We assume it is integers or errors
+          opts = [opts].flatten
+          opts.each do |o|
+            case
+            when o.is_a?(Fixnum)
+              instance.ignore_exit_values << o
+            when o.is_a?(Range)
+              instance.ignore_exit_values += o.to_a
+            when o <= Exception
+              instance.ignore_failures_array << o
```
stevendanna: Would it be good to have an else case here rather than silently ignoring other options?

lib/chef/provisioning/convergence_strategy/ignore_convergence_failure.rb
```diff
@@ -0,0 +1,53 @@
...
+        if opts == true
+          instance.ignore_failures_array << RuntimeError
+        else
+          # We assume it is integers or errors
+          opts = [opts].flatten
+          opts.each do |o|
+            case
+            when o.is_a?(Fixnum)
+              instance.ignore_exit_values << o
+            when o.is_a?(Range)
```
stevendanna: We might want to document that only Ranges, Fixnum's and Exceptions are allowed here.  Would there be any value in allowing anything that responds to `:to_a`?

lib/chef/provisioning/convergence_strategy/ignore_convergence_failure.rb
```diff
@@ -0,0 +1,53 @@
...
+              instance.ignore_exit_values += o.to_a
+            when o <= Exception
+              instance.ignore_failures_array << o
+            end
+          end
+        end
+      end
+
+      def converge(action_handler, machine)
+        begin
```
stevendanna: Minor style point, I don't think you need the begin block since you are rescuing the entire body of the function.

lib/chef/provisioning/convergence_strategy/ignore_convergence_failure.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    # The purpose of this class is to decore the `converge` method with logic to catch any
+    # convergence failure exceptions, log them and then squelch them.  The reason we
+    # need this is to prevent 1 provisioned node's converge failing an entire provisioning
+    # recipe.
+    module IgnoreConvergenceFailure
+
+      attr_accessor :ignore_failures_array, :ignore_exit_values
+
+      # This module is only meant to be extended into instances, not classes or modules
```
stevendanna: It might be useful to have a comment explaining why.  It isn't clear to me why this style of composition is desirable here.

lib/chef/provisioning/convergence_strategy/ignore_convergence_failure.rb
tyler-ball: I dithered on that one.  Ultimately I decided to ignore unknown values so they don't get a runtime error halfway through a provisioning recipe because they specified an unknown option.  Also, I think this is non-intuitive enough that people won't be trying it randomly - they will be reading the `README` and examples.

lib/chef/provisioning/convergence_strategy/ignore_convergence_failure.rb
tyler-ball: I documented it in the [`example in the README`](https://github.com/chef/chef-provisioning/blob/9af761a5d45a4ea8d389df3a3dc5eda3f775edde/README.md#machine-options) - do you think it needs expanded documentation?

lib/chef/provisioning/convergence_strategy/ignore_convergence_failure.rb
tyler-ball: true!

lib/chef/provisioning/convergence_strategy/ignore_convergence_failure.rb
tyler-ball: I don't want to extend it on the `class` level for the `install_sh` convergence strategy because users may have 1 machine with `ignore_failures` and 1 machine without it.  Thats why I extend it on the instance of the convergence_strategy - those are created per-machine.  I'll update the comment.

PR #415, opened 2015-08-13
Run our tests in a matrix of Chef client versions.

randomcamel:
Head-of-master fails on `rake install` for no discernible reason, but the specs pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,31 @@
 sudo: false
 language: ruby
 rvm:
-- 2.1.2
+  - 2.1.2
+
+gemfile:
+  - chef-client-version.gemfile
+
+env:
+  - CHEF_VERSION=master
+  - CHEF_VERSION=12.4.1
+  # - CHEF_VERSION=12.4.0    # we know 12.4.0 doesn't work and specifically exclude it.
+  - CHEF_VERSION=12.3.0
+  - CHEF_VERSION=12.2.1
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -1,10 +1,31 @@
...
+  - CHEF_VERSION=12.1.1
+  - CHEF_VERSION=12.1.0
+
+matrix:
+  # fast_finish: true
+  allow_failures:
+    - env: CHEF_VERSION=master
+
 script:
+  - bundle install --jobs=3 --retry=3
```
tyler-ball: What was the need for 3 retries on this?

bin/generate_driver
```diff
@@ -105,6 +105,7 @@ file prefix("resource/#{resource_name}.rb") do
   content <<-EOS
 # For details on LWRPs, please see the documentation at http://docs.getchef.com/lwrp.html.
 class Chef::Resource::#{camel_name}ExampleResource < Chef::Resource::LWRPBase
+  self.resource_name = "#{resource_name}"
```
tyler-ball: thankyouthankyouthankyou

.travis.yml
randomcamel: I just copied Travis's default `bundle install` command, which (in seeming contravention of the documentation) it was failing to run against a custom gemfile. (I don't actually think we need a custom gemfile, I can just put everything in `Gemfile`, but no need to do that now.)

PR #416, opened 2015-08-14
Add provides statements to providers to avoid chef-client warnings

stevendanna:


tyler-ball:
Future note: we don't need backwards compatibility checks because the required version of cheffish depends on a version of chef which has `provides`.

PR #424, opened 2015-08-24
Specify additional machine_options from the resource attributes

tyler-ball:
This is to support https://github.com/chef/chef-provisioning-aws/pull/299

To make tagging cleaner I wanted to enable users to use `aws_tags` as a resource attribute instead of having to specify it in the `machine_options`.  This brings `machine` in line with other aws resources.

If we like this I'll have to expand it to the `load_balancer` resource as well.  The other option is that we continue having users specify `aws_tags` in the `machine_options` until we refactor away driver.rb

### DIFF
--- a/lib/chef/provider/load_balancer.rb
+++ b/lib/chef/provider/load_balancer.rb
@@ -17,7 +17,7 @@ class Chef
       end
 
       def new_driver
-        @new_driver ||= run_context.chef_metal.driver_for(new_resource.driver)
+        @new_driver ||= run_context.chef_provisioning.driver_for(new_resource.driver)
       end
 
       def chef_managed_entry_store
--- a/lib/chef/provider/machine.rb
+++ b/lib/chef/provider/machine.rb
@@ -147,10 +147,27 @@ class Machine < Chef::Provider::LWRPBase
         end
     }
 
+    # The current use case for this is adding a new attribute `aws_tags` to
+    # the machine resource from the chef-provisioning-aws driver.  Because we
+    # want that attribute to work for the recipe DSL, it needs to be added at the
+    # Chef::Resource::Machine class level and not at an instance level.  Thus we
+    # also need to pull the additional_machine_option_keys (:aws_tags) from the
+    # Chef::Resource::Machine class level.  If you use two drivers (like AWS and
+    # Azure) then all machine instances will still have the `aws_tags` attribute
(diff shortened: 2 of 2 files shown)


lib/chef/provider/load_balancer.rb
```diff
@@ -17,7 +17,7 @@ def whyrun_supported?
       end
 
       def new_driver
-        @new_driver ||= run_context.chef_metal.driver_for(new_resource.driver)
+        @new_driver ||= run_context.chef_provisioning.driver_for(new_resource.driver)
```
randomcamel: _Le roi est mort, vive le roi!_

lib/chef/provider/machine.rb
```diff
@@ -146,6 +146,12 @@ def machine_options(driver)
           result
         end
     }
+    
+    # I'm making the assumption here that you won't be using 2 different drivers
+    # in the same Ruby instance
```
randomcamel: That doesn't feel like a good assumption.

tyler-ball: Yeah, I completely agree.  Here is my reasoning:

To add a new attribute I _must_ do it at the class level (unless someone knows differently).  Or I need to introduce a completely new resource and make the `machine` DSL use my new resource rather than the original resource.  Adapting the `resource_mapping` logic to use the `current_driver` to lookup the correct resource is a _lot_ of work for a system that I don't want to maintain.  Eventually I want users to type `aws_instance` instead of `machine`.

So thats why I decided to add `aws_tags` as an attribute to _all_ `machine`s if the AWS driver is loaded.  I also don't know of any users managing multi-provider cookbooks.

Is this explanation reason enough to keep the code as it is?  Would you like me to update the comment to be more informative?

jkeiser: I don't think there's another way to do it; and as long as the attribute has the word `aws` in it, I think it's OK for now personally. Use it on a vagrant machine and it just won't do anything.

randomcamel: Please update the comment: the way I read it now (which led to my question), it sounds like there's some incompatibility that will happen if two drivers are loaded at the same time.

tyler-ball: Comment updated!

randomcamel:
:shipit:

PR #426, opened 2015-08-26
Major generator/spec/Travis improvements

randomcamel:


bin/generate_driver
```diff
@@ -118,6 +118,16 @@ EOS
 end
 
 # -----------------------------------------
+# machine resource basics.
```
tyler-ball: What does this stuff mean?

spec/spec_helper.rb
```diff
@@ -1,38 +1,16 @@
...
-# a separate helper file that requires the additional dependencies and performs
-# the additional setup, and require it from the spec files that actually need
-# it.
-#
-# The `.rspec` file also contains a few flags that are not defaults but that
-# users commonly want.
-#
-# See http://rubydoc.info/gems/rspec-core/RSpec/Core/Configuration
+require 'simplecov'
+SimpleCov.start do
```
tyler-ball: What are we covering in here?

bin/generate_driver
randomcamel: That's a somewhat WIP thing where I was about to start experimenting with running unit specs on the generated driver. I can remove it for this PR, though probably no one should be looking at this file anyway.

spec/spec_helper.rb
randomcamel: Straight-up `chef-provisioning`.

tyler-ball: But it looks to me like everything between the `do`/`end` blocks is commented out - does it still capture coverage for the specs?

tyler-ball:
:+1:

spec/spec_helper.rb
randomcamel: Yep, the block just allows configuration frobbing, the `start` is the necessary secret sauce to run it.

Issue #433, opened 2015-09-06
ohai_hints should be be created at c:\chef\ohai\hints when provisioning windows nodes


hh:
Looks like we should use ```ChefConfig::Config.platform_specific_path(ohai_hint)``` at  https://github.com/chef/chef-provisioning/blob/master/lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb#L129

Currently ```c:\etc\chef\ohai\hints\ec2.json``` gets created.

```ruby
      # Create the ohai file(s)
      def create_ohai_files(action_handler, machine)
        if convergence_options[:ohai_hints]
          convergence_options[:ohai_hints].each_pair do |hint, data|
            # The location of the ohai hint
            ohai_hint = "/etc/chef/ohai/hints/#{hint}.json"
            machine.write_file(action_handler, ohai_hint, data.to_json, :ensure_dir => true)
          end
        end
      end
```

Ohai looks for this file by checking platform_specific_path:

https://github.com/chef/ohai/blob/master/lib/ohai/config.rb#L35

```ruby
    def self.default_hints_path
      [ ChefConfig::Config.platform_specific_path('/etc/chef/ohai/hints') ]
    end
```

Which uses ```c:\chef\``` platform_specific_hints on windows:

https://github.com/chef/chef/blob/master/chef-config/lib/chef-config/config.rb#L50

```ruby
      if ChefConfig.windows?
        # turns \etc\chef\client.rb and \var\chef\client.rb into C:/chef/client.rb
        if env['SYSTEMDRIVE'] && path[0] == '\\' && path.split('\\')[2] == 'chef'
          path = PathHelper.join(env['SYSTEMDRIVE'], path.split('\\', 3)[2])
        end
      end
```

hh:
If we are provisioning from a linux/osx node to a windows node I'm not quite sure how to do the right thing...

```
[129, 138] in /home/hh/.rvm/gems/ruby-2.2.0/bundler/gems/chef-provisioning-287928054ac6/lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb
   129:             # The location of the ohai hint
   130:             ohai_hint = "/etc/chef/ohai/hints/#{hint}.json"
   131:             # It's in a different path on windows
   132:             ohai_hint = ChefConfig::Config.platform_specific_path(ohai_hint)
   133:             byebug
=> 134:             machine.write_file(action_handler, ohai_hint, data.to_json, :ensure_dir => true)
   135:           end
[...]

Issue #431, opened 2015-09-06
omnibus commandline over winrm  is too long


hh:
When provisioning windows boxes with the winrm omnibus installer we get an error ```
STDERR: The command line is too long. 
```

# The tail output from the chef-client run on the provisioner:

```
    - create directory \etc\chef\ohai\hints on windows1
    - write file /etc/chef/ohai/hints/ec2.json on windows1
    - write file C:\chef\client.rb on windows1
    ================================================================
    Error executing action `converge` on resource 'machine[windows1]'
    ================================================================
    
    RuntimeError
    ------------
    Error: command '
    $chef_omnibus_root = "$env:systemdrive\opscode\chef"
    $msi = "$env:TEMP\chef-latest.msi"
    $chef_metadata_url = "https://www.chef.io/chef/metadata?p=windows&m=x86_64&pv=2008r2&v=latest"
    $pretty_version = "always install latest version"
    $version = "latest"
    
    Function Check-UpdateChef($root, $version) {
      if (-Not (Test-Path $root)) { return $true }
      elseif ("$version" -eq "true") { return $false }
      elseif ("$version" -eq "latest") { return $true }
    
      Try { $chef_version = Get-Content $root\version-manifest.txt | select-object -1}
      Catch {
        Try { $chef_version = (& $root\bin\chef-solo.bat -v).split(" ", 2)[1] }
        Catch { $chef_version = "" }
      }
    
      if ($chef_version.StartsWith($version)) { return $false }
      else { return $true }
    }
    
    Function Get-ChefMetadata($url) {
      Try { $response = ($c = Make-WebClient).DownloadString($url) }
      Finally { if ($c -ne $null) { $c.Dispose() } }
    
      $md = ConvertFrom-StringData $response.Replace("`t", "=")
      return @($md.url, $md.md5)
    }
    
    Function Get-MD5Sum($src) {
      Try {
        $c = New-Object -TypeName System.Security.Cryptography.MD5CryptoServiceProvider
        $bytes = $c.ComputeHash(($in = (Get-Item $src).OpenRead()))
        return ([System.BitConverter]::ToString($bytes)).Replace("-", "").ToLower()
      } Finally { if (($c -ne $null) -and ($c.GetType().GetMethod("Dispose") -ne $null)) { $c.Dispose() }; if ($in -ne $null) { $in.Dispose() } }
    }
    
    Function Download-Chef($url, $md5, $dst) {
      Try {
        Log "Downloading package from $url"
[...]

PR #434, opened 2015-09-06
use platform_specific_path(ohai_hint) on windows

hh:
There is no '/etc' on windows. Should fix chef/chef-provisioning#433

### DIFF
--- a/lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb
+++ b/lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb
@@ -127,6 +127,8 @@ module Provisioning
           convergence_options[:ohai_hints].each_pair do |hint, data|
             # The location of the ohai hint
             ohai_hint = "/etc/chef/ohai/hints/#{hint}.json"
+            # It's in a different path on windows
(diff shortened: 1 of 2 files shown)


hh:
Hhmmm... this platform_specific_hint should actually be based on the target OS, no t the current one

hh:
https://github.com/chef/chef-provisioning/issues/431

In debugging we needed to detect is the machine we were targeting was windows.
I hardcoded the drive to 'c:' bet we should probably probe for it.


```
[133, 142] in /home/hh/.rvm/gems/ruby-2.2.0/bundler/gems/chef-provisioning-287928054ac6/lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb
   133:             if  machine.class == Chef::Provisioning::Machine::WindowsMachine
   134:               # we may have a different system drive, but we'd need to ask for it
   135:               ohai_hint = ChefConfig::PathHelper.join('c:', ohai_hint.split('/', 3)[2..-1])
   136:             end
   137:             byebug
=> 138:             machine.write_file(action_handler, ohai_hint, data.to_json, :ensure_dir => true)
   139:           end
   140:         end
   141:       end
   142: 
(byebug) ohai_hint
"c:/chef/ohai/hints/ec2.json"
```

PR #436, opened 2015-09-09
Adding additional resource attributes for load_balancer and machine_image

tyler-ball:
This is an extension of https://github.com/chef/chef-provisioning/pull/424

\cc @jkeiser @randomcamel


lib/chef/provider/machine.rb
```diff
@@ -157,7 +157,7 @@ def machine_options(driver)
     # DSL and will pass `:aws_tags` in on the machine_options.  They can simply
     # be ignored by the Azure driver.
     (self.class.additional_machine_option_keys || []).each do |k|
-      configs << { k => new_resource.send(k)}
+      configs << { k => new_resource.send(k)} if new_resource.send(k)
```
tyler-ball: Adding this means we will converge tags on an instance that already exists

randomcamel:
:+1:

PR #437, opened 2015-09-09
Adding a CONTRIBUTING document

tyler-ball:
Which describes acceptance criteria for all Chef Provisioning resources.

\cc @randomcamel @jkeiser @charlesjohnson 

- [ ] https://github.com/chef/chef-provisioning-aws/pull/316
- [x] https://github.com/chef/chef-provisioning-azure/pull/38
- [ ] https://github.com/chef/chef-provisioning-docker/pull/64
- [x] https://github.com/chef/chef-provisioning-fog/pull/121
- [ ] https://github.com/chef/chef-provisioning-ssh/pull/24
- [ ] https://github.com/chef/chef-provisioning-vagrant/pull/30

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,20 @@
+# Contributing to Chef Provisioning
+
+## Resource/Provider Acceptance Criteria
+
+There is a common set of acceptance criteria for all Chef Provisioning resources.
+This is the minimum set of acceptance criteria - individual projects may expand
+these.
+
+In order to be complete, resources in this library must have the following properties:
+1. The resource must have fully functional `:create` and `:destroy` actions.
+  1. If the object specified by the resource already exists, and the `:create` action is called, then the `:create` action must update all available attributes for the specified object.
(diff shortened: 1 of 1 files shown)


randomcamel:
:+1: Slick.

stuartpreston:
Excuse my naivety - why must **all** Chef Provisioning resources implement :create and :destroy actions?
What if 'create' is the wrong verb (e.g. deploy a CF template, or in my case an Azure Resource Manager one)

tyler-ball:
@stuartpreston I think `create` is actually the wrong word - I would much more prefer `converge` as the default action.  As you say, a Chef Resource may model an object which isn't actually created just interacted with.  But the use of `create` as the default action goes wayyyyyyyy back.

If Chef Provisioning Azure should use some other actions, then we could update the CONTRIBUTING.md document there to say what the minimum set of actions the resources should respond to.

The goal here is not to force all Provisioning Resources to conform to a too-simple API, but rather to define a set of criteria so Resource authors know when their Resource is complete.

PR #439, opened 2015-09-10
Added missing arguments to call of chef install shell script.

tarak:
While trying to provision a machine using chefdk I wondered why the usual chef-client package was installed instead. The `install_sh_arguments` is not passed to the `install_sh_path` call in https://github.com/chef/chef-provisioning/blob/master/lib/chef/provisioning/convergence_strategy/install_sh.rb#L45 - This PR wraps the call into a string, hence to single quotes inside the double quotes.


tyler-ball:
Thanks for finding this @tarak !

irvingpop:
I just noticed that this fix doesn't work -  because the `chef-install.sh` file generated by `mixlib-install` doesn't appear to take any arguments on the command line.   I've created this PR to fix that:  https://github.com/chef/chef-provisioning/pull/452

PR #445, opened 2015-09-16
Custom port forwards.

causton81:
In my environment I have a private cloud where I can spin up new VMs with no public network access nor access to the company git repos.  With the changes in this PR I can run a squid proxy on my workstation and forward that out to each VM for http proxy. I also forward out a remote port for our git server.

These changes work pretty well for me, but I'm curious what you all think.

### DIFF
--- a/lib/chef/provisioning/transport/ssh.rb
+++ b/lib/chef/provisioning/transport/ssh.rb
@@ -42,6 +42,7 @@ module Provisioning
         @ssh_options = ssh_options
         @options = options
         @config = global_config
+        @remote_forwards = ssh_options.delete(:remote_forwards) { Array.new }
       end
 
       attr_reader :host
@@ -51,12 +52,25 @@ module Provisioning
       attr_reader :config
 
       def execute(command, execute_options = {})
-        Chef::Log.info("Executing #{options[:prefix]}#{command} on #{username}@#{host}")
+        Chef::Log.info("#{self.object_id} Executing #{options[:prefix]}#{command} on #{username}@#{host}")
         stdout = ''
         stderr = ''
         exitstatus = nil
         session # grab session outside timeout, it has its own timeout
+
         with_execute_timeout(execute_options) do
+          @remote_forwards.each do |forward_info|
+              # -R flag to openssh client allows optional :remote_host and
+              # requires the other values so let's do that too.
+              remote_host = forward_info.fetch(:remote_host, 'localhost')
+              remote_port = forward_info.fetch(:remote_port)
+              local_host = forward_info.fetch(:local_host)
+              local_port = forward_info.fetch(:local_port)
+
+              actual_port, actual_host = forward_port(local_port, local_host, remote_port, remote_host)
+              Chef::Log.info("#{host} forwarded remote #{actual_host}:#{actual_port} to local #{local_host}:#{local_port}")
+          end
+
           channel = session.open_channel do |channel|
             # Enable PTY unless otherwise specified, some instances require this
             unless options[:ssh_pty_enable] == false
@@ -85,6 +99,20 @@ module Provisioning
           end
 
           channel.wait
+
+          @remote_forwards.each do |forward_info|
+              # -R flag to openssh client allows optional :remote_host and
+              # requires the other values so let's do that too.
+              remote_host = forward_info.fetch(:remote_host, 'localhost')
+              remote_port = forward_info.fetch(:remote_port)
+              local_host = forward_info.fetch(:local_host)
+              local_port = forward_info.fetch(:local_port)
+
+              session.forward.cancel_remote(remote_port, remote_host)
+              session.loop { session.forward.active_remotes.include?([remote_port, remote_host]) }
+
+              Chef::Log.info("#{host} canceled remote forward #{remote_host}:#{remote_port}")
+          end
         end
 
         Chef::Log.info("Completed #{command} on #{username}@#{host}: exit status #{exitstatus}")
@@ -156,15 +184,17 @@ module Provisioning
           end
           uri.host = host
           uri.port = port
+          Chef::Log.info("Port forwarded: local URL #{local_url} is available to #{self.host} as #{uri.to_s} for the duration of this SSH connection.")
+        else
+          Chef::Log.info("#{host} not forwarding non-local #{local_url}")
         end
-        Chef::Log.info("Port forwarded: local URL #{local_url} is available to #{self.host} as #{uri.to_s} for the duration of this SSH connection.")
         uri.to_s
       end
 
       def disconnect
         if @session
           begin
-            Chef::Log.debug("Closing SSH session on #{username}@#{host}")
+            Chef::Log.info("Closing SSH session on #{username}@#{host}")
             @session.close
           rescue
           ensure
@@ -333,7 +363,7 @@ module Provisioning
 
           actual_remote_port, actual_remote_host = session.forward.active_remote_destinations[[local_port, local_host]]
           if !actual_remote_port
-            Chef::Log.debug("Forwarding local server #{local_host}:#{local_port} to #{username}@#{self.host}")
+            Chef::Log.info("Forwarding local server #{local_host}:#{local_port} to #{username}@#{self.host}")
 
             session.forward.remote(local_port, local_host, remote_port, remote_host) do |new_remote_port, new_remote_host|
 							actual_remote_host = new_remote_host
--- a/README.md
+++ b/README.md
@@ -198,7 +198,21 @@ with_machine_options({
   ssh_options: { # a list of options to Net::SSH.start
     :auth_methods => [ 'publickey' ], # DEFAULT
     :keys_only => true, # DEFAULT
-    :host_key_alias => "#{instance.id}.AWS" # DEFAULT
+    :host_key_alias => "#{instance.id}.AWS", # DEFAULT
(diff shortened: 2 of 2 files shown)


chef-supermarket:
Hi. I am an automated pull request bot named Curry. There are commits in this pull request whose authors are not yet authorized to contribute to Chef Software, Inc. projects or are using a non-GitHub verified email address. To become authorized to contribute, you will need to sign the Contributor License Agreement (CLA) as an individual or on behalf of your company. [You can read more on Chef's blog.](https://www.getchef.com/blog/2014/06/23/changes-to-the-contributor-license-agreement-process)

## Non-GitHub Verified Committers

There are 1 commit author(s) whose commits are authored by a non-GitHub verified email address. Chef will have to manually verify that they are authorized to contribute.

causton81:
Not sure what the CLA thing is about, I signed it long ago: https://supermarket.chef.io/users/causton81

lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb
```diff
@@ -31,6 +31,30 @@ def setup_convergence(action_handler, machine)
...
+            if path_or_config.class <= Hash
+                normal_path = path_or_config[:normal_path]
+                options.merge! path_or_config
+            else
+                normal_path = path_or_config
+            end
+
+            remote_url = machine.make_url_available_to_remote(local_url, options)
+
+            parts = normal_path.split('/')
```
tyler-ball: Hey @causton81, I am confused - what are lines 46-56 doing?  Adding the forwarded ports to the machine_spec?

lib/chef/provisioning/transport/ssh.rb
```diff
@@ -91,6 +91,9 @@ def execute(command, execute_options = {})
         Chef::Log.debug("Stdout was:\n#{stdout}") if stdout != '' && !options[:stream] && !options[:stream_stdout] && config[:log_level] != :debug
         Chef::Log.info("Stderr was:\n#{stderr}") if stderr != '' && !options[:stream] && !options[:stream_stderr] && config[:log_level] != :debug
         SSHResult.new(command, execute_options, stdout, stderr, exitstatus)
+      rescue Exception => e
+        Chef::Log.error("Problem executing command on #{host} #{e}")
```
tyler-ball: Did this help with debugging?  I've started doing this pattern myself so I make sure the message is stored in a log file and not just on STDOUT

tyler-ball:
I'm :+1: except for 1 issues - this needs documentation.  Can you update https://github.com/chef/chef-provisioning/blob/master/README.md#machine-options with examples and explanations of all the available keys/values that can be provided to `remote_forwards`?

lib/chef/provisioning/convergence_strategy/precreate_chef_objects.rb
```diff
@@ -31,6 +31,30 @@ def setup_convergence(action_handler, machine)
         chef_server_url = chef_server[:chef_server_url]
         chef_server_url = machine.make_url_available_to_remote(chef_server_url)
 
+        convergence_options[:remote_forwards].each_pair do |local_url, path_or_config|
```
tyler-ball: Because this only pertains to SSH (winrm does not support port forwarding) we should make this part of the `ssh_options` machine_option and move it into the SSH transport class.

Note:  I think the `ssh_options` are passed to the SSH transport class as the `options` parameter

causton81:
This all sounds good to me. I will try to get something in late next week.

causton81:
This probably needs some log cleanup, but I left it this way to illustrate why I chose this implementation. At least the fog driver creates multiple ssh transports for the same host and never disconnects them explicitly. This causes problems if each new transport tries to forwards ports. The worst problem is a transport that has not yet been garbage collected and its IO loop is not being processed; this looks fine on the remote (the ports are listening) but if you try to use the ports they just hang (IO loop not being processed). Sometimes the port forwards will fail immediately when net-ssh tries to create them; my guess is this is something to do with TIME_WAIT in TCP.

Since there are no guarantees on how ssh transports are used by drivers, the best option seems to be to forward the ports just before executing a command and then immediately canceling the forwards after the command completes.

causton81:
I forgot to mention that there is a race condition in net-ssh that I am working to fix upstream here: https://github.com/net-ssh/net-ssh/pull/280

May want to hold off on merging this until then.

lib/chef/provisioning/transport/ssh.rb
```diff
@@ -85,6 +99,20 @@ def execute(command, execute_options = {})
...
+          @remote_forwards.each do |forward_info|
+              # -R flag to openssh client allows optional :remote_host and
+              # requires the other values so let's do that too.
+              remote_host = forward_info.fetch(:remote_host, 'localhost')
+              remote_port = forward_info.fetch(:remote_port)
+              local_host = forward_info.fetch(:local_host)
+              local_port = forward_info.fetch(:local_port)
+
+              session.forward.cancel_remote(remote_port, remote_host)
+              session.loop { session.forward.active_remotes.include?([remote_port, remote_host]) }
```
tyler-ball: It seems like we want some kind of timeout on this to ensure we don't get stuck here - does `.loop` already have a default timeout or maximum tries?

tyler-ball:
Have you tested what happens if the user specifies a `remote_forward` that conflicts with something setup automatically by the `make_url_available_to_remote` method?  The only time I know we use that is when the user runs a provisioning recipe with `chef-client -z` and we automatically forward the local chef-zero server to the remote instance.  I'm pretty sure the default port is 8889.

Can you test that and see what happens?  If it is a nasty error we will need to add some checking.

Otherwise I'll wait on your feedback from https://github.com/net-ssh/net-ssh/pull/280

causton81:
I'll give it a test. net-ssh upstream has moved to 3.x but chef is still pinned to 2.x.  Would chef bump to 3.x or will we need to get my PR backported to net-ssh 2.10?

tyler-ball:
@causton81 Good question - @lamont-granquist @danielsdeleo @jkeiser do you know any reason we couldn't upgrade to net-ssh 3.x when https://github.com/net-ssh/net-ssh/pull/280 is merged?

The [Changelog](https://github.com/net-ssh/net-ssh/blob/master/CHANGES.txt) lists a breaking change for `exec!`  I can't find any uses of that in the Chef codebase.  `Major version change because of dropping of ruby 1.9`.

pburkholder:
I have some notes on testing net-ssh 3.0.1 with the channel-close patch here: https://gist.github.com/pburkholder/05455d33d4a9f83a54b6

causton81:
@pburkholder, very cool using `tc` to demonstrate the issue.

lamont-granquist:
only issue is that probably everyone pins net-ssh with '~> 2.x' so you'll have to make sure to bump all the consumers at the same time or you'll get into gem conflicts.  this repo and chef/chef at the very least will need synchronized bumps.  i'd also advise using '>= 2.0, < 4.0' instead of '~> 3.x' everywhere that you can in order to favor compatibility, unless you really need to drop support for 2.x

danielsdeleo:
:+1: on what Lamont said, though we won't get the benefits of upgrading until everything that gets bundled together upgrades. So that might be an argument for using `~> 3.x` and being aggressive about fixing issues we find.

lamont-granquist:
Yeah if you go that route just be aware that the first time you commit '~> 3.x' that you most likely break the world and then buy yourself a yak shave hunting down repos that need to get bumped.   And I suspect its probably better to pin it in one place (probably here) and then let chef/chef float on both, since that'll get you the pinning you desire while also minimizing the amount of tooling that instantly breaks.

causton81:
FYI https://github.com/net-ssh/net-ssh/tree/2-9-release

tyler-ball:
@causton81 Thanks!  It looks like the fix got backported to the 2.9 release family - we should be able to take 2.9.4

lib/chef/provisioning/transport/ssh.rb
causton81: Hmm, I think the whole thing is inside `with_execute_timeout` so does that handle your concern?

causton81:
Yeah, looking forward to 2.9.4 getting released. I guess the log statements need to be cleaned up if you are okay with the rest of the logic?

tyler-ball:
@causton81 We have a build of the ChefDK with only net-ssh 3.0.2 in it.  Would you mind installing with the instructions [here](https://github.com/chef/chef-dk#pre-release-candidates) and see if that fixes your race condition?

causton81:
Sure, I'll give it a try tomorrow hopefully :D

Chris Auston
cel: 703-209-9668
causton81@gmail.com

causton81:
Does the current version of ChefDK include this PR? It seems not:
```
ArgumentError: machine_batch[batch 1] (oss-cluster::default line 321) had an error: ArgumentError: invalid option(s): remote_forwards
/opt/chefdk/embedded/lib/ruby/gems/2.1.0/gems/net-ssh-3.0.2/lib/net/ssh.rb:197:in `start'
```

For me the race condition was most obvious using the remote port forwards setup by this PR.

causton81:
@tyler-ball I installed the chef-dk build you linked and updated chef-provisioning to this PR. It worked very well for me.

lib/chef/provisioning/transport/ssh.rb
tyler-ball: :+1:

PR #459, opened 2015-10-08
Pin mixlib-install more strictly.

sersut:
We are doing some work on mixlib-install to extend its functionality and we will be making some compat breaking changes. E.g. chef/mixlib-install#11

We would like to pin it more strictly here until we get the mixlib-install to 1.0.0 stage.


jkeiser:
:+1: that's fine. semver isn't guaranteed pre-1.0 anyway.

sersut:
Thanks @jkeiser

PR #466, opened 2015-10-20
Making available? timeout use provided ssh_options, fixes #362

### LINKED ISSUE #362: SSH available timeout needs to be configurable
[Here](https://github.com/chef/chef-provisioning/blob/master/lib/chef/provisioning/transport/ssh.rb#L171) in the `available?` method we hardcode the SSH timeout to 10 seconds - this causes errors where we will wait the full 120 seconds for an instance to boot (even though it booted long ago) because its SSH is responding suuuuuuper slow.

I think we should remove the timeout on that line and instead let the [ssh_options](https://github.com/chef/chef-provisioning/blob/master/lib/chef/provisioning/transport/ssh.rb#L187) passed in dictate the timeout:

```ruby
machine "test" do
  machine_options :ssh_options => {
    :timeout => 30
  }
end
```

tyler-ball:
So super slow connections (like CI) can still work, fixes https://github.com/chef/chef-provisioning/issues/362

\cc @randomcamel @jkeiser

### DIFF
--- a/lib/chef/provisioning/transport/ssh.rb
+++ b/lib/chef/provisioning/transport/ssh.rb
@@ -174,8 +174,8 @@ module Provisioning
       end
 
       def available?
-        # If you can't pwd within 10 seconds, you can't pwd
-        execute('pwd', :timeout => 10)
+        timeout = ssh_options[:timeout] || 10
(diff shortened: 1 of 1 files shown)


randomcamel:
:+1:

PR #478, opened 2015-12-09
Require ResourceBuilder file before monkeypatching to ensure it is already defined

tyler-ball:
When monkeypatching the ResourceBuilder class we first need to ensure the primary instance is already defined.  Otherwise our monkeypatching becomes the primary instance and the real instance is never defined.

\cc @jkeiser @smith @randomcamel @jaym @btm

### DIFF
--- a/lib/chef/provisioning/driver.rb
+++ b/lib/chef/provisioning/driver.rb
@@ -311,12 +311,12 @@ end
 
 # In chef-provisioning we don't perform resource cloning
 # This fixes resource cloning when the ResourceBuilder is present
-
+require 'chef/resource_builder' unless defined?(Chef::ResourceBuilder)
 class Chef
   class ResourceBuilder
     if defined?(:prior_resource)
       def prior_resource
-        # NOOP
+        nil
       end
     end
   end


jaym:
@lamont-granquist mentioned that lazy loading was required... I'm cant remember why. Thoughts?

lamont-granquist:
lazy loading is required in the code inside chef/chef, i've got no idea about this use here....

the bigger problem with this is that monkeypatching is horrible and the chef/chef API needs to be extended so that it has a public interface that chef-provisioning can use instead of this....

since we're blocking releases though :+1:

tyler-ball:
@lamont-granquist why is lazy loading required inside chef/chef?  Because chef-provisioning is 'just another chef recipe' :tm: we may encounter the same issue.

And I totally agree with you about modifying the interface.  I can open up a separate chef/chef PR with that

lamont-granquist:
it blows up horribly in chef/chef due to circular require trees.  chef/chef isn't recipe code.  if your rspec tests pass here (and you've got halfway decent coverage) you are probably fine since you're externally requiring it.

in fact the lazy require in chef/chef is probably the fix for the circular require tree you'd get if that wasn't there...

adamedx:
Thanks for jumping on this @tyler-ball.

PR #485, opened 2016-01-14
Updating to the latest release of net-ssh to consume net-ssh/net-ssh#280

tyler-ball:
All the products in the ChefDK are getting updated to 3.0.2


lamont-granquist:
:+1:

Issue #443, opened 2015-09-16
SchTask Support for ConvergenceStrategy::InstallMsi.converge


hh:
It's a dependency for getting machine_image supported for run_lists that include wsus updates and other tasks required to be run on \LocalMachine.

chef/mixlib-install#9 should get us to a point where MixLib install support SchTask
by creating a [install_command](https://github.com/chef/chef-provisioning/blob/master/lib/chef/provisioning/convergence_strategy/install_msi.rb#L39) but I think we'll need make changes to  ConvergenceStrategy::InstallMsi.converge to support creation / usage of a chef-client-schtask

hh:
Related if we want a pure SchTask based install and run:

chef/mixlib-install/#9

Background:

chef/chef-provisioning#409
chef/chef-provisioning#410

hh:
We need updates applied, and they can only be applied successfully via schtask.
This is a spike that works, but I'd definitely feeling the pain of our lack of schtask provisioning to do this properly. (It sorta works, I get a 401 authentication fails after the wsus-updates apply, but that a different issue) @adamedx @tyler-ball @mwrock  ?

```ruby
require 'chef/provisioning/aws_driver'
require 'pry-byebug'
current_dir = File.dirname(__FILE__) # to get relative files... encrypted databag

with_chef_server "https://api.chef.io/organizations/#{ENV['CHEF_ORG']}",
                 :client_name => Chef::Config[:node_name],
                 :signing_key_filename => Chef::Config[:client_key]

with_machine_options bootstrap_options: {
                       instance_type: 'c3.2xlarge',
                       subnet_id: 'subnet-6021b417',
                       security_group_ids: [
                         'sg-689c5a0c',
                         'sg-6f9c5a0b',
                         'sg-0a9a5c6e'
                       ]
                     },
                     aws_tags: {
                       :created_by => 'chris',
                       :please_destroy => 'true'
                     },
                     transport_address_location: :private_ip,
                     image_id: 'ami-7bc3e04b' # aws-marketplace/CIS Microsoft Windows Server 2012 R2 Benchmark v1.1.0-26bb465c-ce26-4da9-afb8-040b2f8c9a7f-ami-7a88f312.2

machine_name = 'base-2012-hardened-xv'

m = machine "#{machine_name}" do
  action :allocate
end

# converge a machine with chef-client::task
machine "#{machine_name}" do
  #action :nothing
  action :converge
  attribute %w{ client_client task frequency}, 'once'
[...]

PR #488, opened 2016-01-28
WinRM::CommandExecutor support

### LINKED ISSUE #487: WinRM::CommandExecutor support

hh:
Should fix #487 though there is probably more to due to bring in support for running via schtask.

#443 and chef/chef-provisioning-aws#324


monkey1016:
Would be great if this could be merged.  WinRM provisioning was failing without this change for me.

tyler-ball:
Newer code implemented in https://github.com/chef/chef-provisioning/pull/543

PR #496, opened 2016-02-01
Fix Provisioning with Cheffish 1.x

jkeiser:
The code used to fix provisioning for Cheffish 2.x broke `with_chef_server` in Cheffish 1.x, which did that by using the `machine` resource as the ChefNode provider's `new_resource` (since they have the same attributes). This reverts us to using the provider instead of the resource to take advantage of that.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -39,11 +39,11 @@ namespace :driver do
   end
 end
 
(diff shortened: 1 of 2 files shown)


jaym:
:+1:

PR #515, opened 2016-04-27
Add support for custom :install_sh_url

### LINKED ISSUE #474: Flexible install_command. Do not assume internet acces.
Currently chef-provisioning assume we have internet access to install chef. It would be super awesome if we could customize the `install_command` coming from here https://github.com/chef/chef-provisioning/blob/master/lib/chef/provisioning/convergence_strategy/install_sh.rb#L44

SIGUSR2:
Fixes #474

### DIFF
--- a/lib/chef/provisioning/convergence_strategy/install_sh.rb
+++ b/lib/chef/provisioning/convergence_strategy/install_sh.rb
@@ -37,6 +37,8 @@ module Provisioning
           opts["https_proxy"] = convergence_options[:bootstrap_proxy]
         end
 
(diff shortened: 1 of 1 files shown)


tyler-ball:
Thanks @SIGUSR2!

One note about this commit - I don't think the convergence options are persisted by default. Meaning, you will need to specify `install_sh_url` in the resource whenever the `converge` action will be performed.

In the future if this is messing people up, we could persist this setting in the `machine_spec`. But for now, this looks great!

tyler-ball:
I was going to update the README when merging this, but I see this PR just brings reality in line with our documentation. Thanks!

SIGUSR2:
Thank you!

djapg:
Thank You!!

PR #518, opened 2016-05-07
install_sh and install_cached convergence strategies trigger chef-client with -c flag

poliva83:
Using :client_rb_path convergence option make install_sh and install_cached convergence strategies trigger chef-client using -c flag pointing at this location where client.rb was created.


irvingpop:
👍 looks super useful

PR #541, opened 2016-08-28
Allow machine_execute resource to run commands that last longer then 15 minutes

poliva83:
Add :timeout attribute to machine_execute resource that will set :timeout option in Chef::Provisioning::Transport.


poliva83:
Thanks for merging @tyler-ball

PR #543, opened 2016-09-02
Use Winrm v2 and let winrm-fs do file uploads

mwrock:
This PR accomplishes 3 things:
* Updates WinRM gem usage to 2.0 API
* Replaces file upload code with wintm-fs (same code used by test-kitchen, inspec and vagrant) with its 1.0 implementation its orders of mafnitude faster the the former implementation and no longer constrained to 8k chunks.
* Use one powershell based shell throughout an entire converge. This is simpler, faster and follows the same approach as test-kitchen.

### DIFF
--- a/lib/chef/provisioning/convergence_strategy/install_msi.rb
+++ b/lib/chef/provisioning/convergence_strategy/install_msi.rb
@@ -41,8 +41,7 @@ module Provisioning
 
         action_handler.open_stream(machine.node['name']) do |stdout|
           action_handler.open_stream(machine.node['name']) do |stderr|
-            machine.execute(action_handler, "powershell.exe -ExecutionPolicy Unrestricted -NoProfile \"& \"\"#{convergence_options[:install_script_path]}\"\"\"",
-              :raw => true,
+            machine.execute(action_handler, "& \"#{convergence_options[:install_script_path]}\"",
               :stream_stdout => stdout,
               :stream_stderr => stderr)
           end
@@ -54,10 +53,11 @@ module Provisioning
 
         action_handler.open_stream(machine.node['name']) do |stdout|
           action_handler.open_stream(machine.node['name']) do |stderr|
-            command_line = "chef-client"
+            # We just installed chef in this shell so refresh PATH from System.Environment
+            command_line = "$env:path = [System.Environment]::GetEnvironmentVariable('PATH', 'MACHINE');"
+            command_line << "chef-client"
             command_line << " -l #{config[:log_level].to_s}" if config[:log_level]
             machine.execute(action_handler, command_line,
-              :raw => true,
               :stream_stdout => stdout,
               :stream_stderr => stderr,
               :timeout => @chef_client_timeout)
--- a/lib/chef/provisioning/transport/winrm.rb
+++ b/lib/chef/provisioning/transport/winrm.rb
@@ -22,27 +22,24 @@ module Provisioning
       # The actual connection is made as ::WinRM::WinRMWebService.new(endpoint, type, options)
       #
       def initialize(endpoint, type, options, global_config)
-        @endpoint = endpoint
-        @type = type
         @options = options
+        @options[:endpoint] = endpoint
+        @options[:transport] = type
+
+        # WinRM v2 switched from :pass to :password
+        # we accept either to avoid having to update every driver
+        @options[:password] = @options[:password] || @options[:pass]
+        
         @config = global_config
       end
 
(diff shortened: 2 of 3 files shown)


mwrock:
No longer WIP and tested using chef-provisioning-vsphere

mwrock:
cc @tyler-ball @jkeiser

tyler-ball:
Thanks @mwrock !

lib/chef/provisioning/transport/winrm.rb
```diff
@@ -22,27 +22,24 @@ class WinRM < Chef::Provisioning::Transport
...
         output = with_execute_timeout(execute_options) do
-          session.set_timeout(execute_timeout(execute_options))
-          command_executor = ::WinRM::CommandExecutor.new(session)
           block = Proc.new { |stdout, stderr| stream_chunk(execute_options, stdout, stderr) }
-          if execute_options[:raw]
-            command_executor.run_cmd(command, &block)
-          else
-            command_executor.run_powershell_script(command, &block)
-          end
+          session.run(command, &block)
```
hh: @mwrock @tyler-ball we need to default to session.run_powershell_script and only use session.run_cmd if raw options are passed the ```:raw``` options
/cc @ajeba99

hh: https://github.com/chef/chef-provisioning/pull/488/files

lib/chef/provisioning/convergence_strategy/install_msi.rb
```diff
@@ -41,8 +41,7 @@ def setup_convergence(action_handler, machine)
 
         action_handler.open_stream(machine.node['name']) do |stdout|
           action_handler.open_stream(machine.node['name']) do |stderr|
-            machine.execute(action_handler, "powershell.exe -ExecutionPolicy Unrestricted -NoProfile \"& \"\"#{convergence_options[:install_script_path]}\"\"\"",
```
hh: @mwrock @tyler-ball this is a must have! We need ExecutionPolicy Unrestricted in order to run powershell
/cc @ajeba99

lib/chef/provisioning/transport/winrm.rb
```diff
@@ -114,9 +99,15 @@ def make_url_available_to_remote(local_url)
 
       def session
         @session ||= begin
-          require 'kconv' # Really, people? *sigh*
           require 'winrm'
-          ::WinRM::WinRMWebService.new(endpoint, type, options)
+          ::WinRM::Connection.new(options).shell(:powershell)
```
hh: Setting shell to powershell here may get rid of the need for the :raw block above

mwrock:
Hi @hh can you provide an example of what does not work? We made the breaking change in 2.0 to no longer support `:raw` and now all transport commands use a powershell shell.

hh:
It's not so much around the :raw, it's more around the ExecutionPolicy


In https://github.com/chef/chef-provisioning/pull/543/files/45e94cc3dbd716c9dc6e380dbd94adf9aa15b28e#diff-72ecfcec074b5e5bdfc04d91922a92fcL44 we lost:

```
   machine.execute(action_handler, "powershell.exe -ExecutionPolicy Unrestricted -NoProfile \"& \"\"#{convergence_options[:install_script_path]}\"\"\"",
```

Which since chef-dk 0.18.26 results in:

```
    - write file C:\chef\install.ps1 on base-2012r2-2016-08-01
    [base-2012r2-2016-08-01] File C:\chef\install.ps1 cannot be loaded because running scripts is disabled on this system. For more information, see about_Execution_Policies at http://go.microsoft.com/fwlink/?LinkID=135170.
                             At line:1 char:3
                             + & "C:\chef\install.ps1"
                             +   ~~~~~~~~~~~~~~~~~~~~~
                                 + CategoryInfo          : SecurityError: (:) [], PSSecurityException
                                 + FullyQualifiedErrorId : UnauthorizedAccess


```

hh:
#550 should fix it

mwrock:
Ah I see. So the big difference with powershell in winrm v2's `:powershell` shell is that it takes you directly to a powershell session and does not go through a cmd.exe process. So downfall of calling `powershell.exe` from a powershell shell is you are creating another powershell process from a powershell process. There are 2 ways to work around this scenario:

1. Make sure your base images set a reasonable execution timeout
2. Call `Set-ExecutionPolicy unrestricted` as the first command. However this will persist

ajeba99:
Added #551

PR #550, opened 2016-10-17
ExecutionPolicy => Unrestricted for install.ps1

hh:


hh:
@mwrock @tyler-ball This is an obvious change, not worth a DCO (I haven't set that up yet)

epsilon-jpage:
:+1: LGTM

mwrock:
I'm 👎 on this since it opens a new powershell.exe process from an existing powershell session. I'd simply call `Set-ExecutionPolicy -Scope Process -ExecutionPolicy Unrestricted` instead.

mwrock:
Another thought is to add winrm-elevated to the dependencies and use an `:elevated` shell instead of `:powershell`. That could fix other things too that don't work over "normal" winrm.

epsilon-jpage:
@mwrock In your change what was the original reasoning why the Powershell window didn't have elevated privileges before?  Was it for security reasons?

mwrock:
@epsilon-jpage it was mainly to make the elevated shell "opt-in" since it adds new and additional functionality that could potentially break especially since winrm-elevated is fairly new.

epsilon-jpage:
@mwrock It sounds like what you're asking for is additional functionality instead of fixing the existing install.ps1 script.  As of now the install.ps1 script is failing to run because it doesn't have have the necessary execution policy.  I would also like to have, at some point, some functionality to control elevated levels of winrm but that's not what was put in #543.  What you're asking for is an enhancement.

ajeba99:
Added #551

Saburesh07:

PR #551, opened 2016-10-18
Set WinRM to Elevated Shell

ajeba99:
In the spirit of time, saving @HH from creating a DCO, and fixing this install.ps1 issue I am submitting this.  

All credit goes to @HH as this isn't my arena.

Although I don't know how to set the timeout on the session / exeuction.

Signed-off-by : Jim Page <ajeba99@gmail.com>


ajeba99:
Signed-off-by: Jim Page <ajeba99@gmail.com>

PR #553, opened 2016-11-05
Add a section about ACLs in README.me

elthariel:
I wasted quite some time trying to understand how to fix properly the `chef_client[mymachine] had an error: Net::HTTPServerException: 403 "Forbidden"` error.

I stumbled accross the knife edit groups/admin.json solution before I was able to find the (outdated) blog post which put me on the way of the cleaner solution. I think we should rather document this in an obvious place

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,6 +293,20 @@ machine 'mario' do
 end
 ```
 
+### Configuring ACLs
+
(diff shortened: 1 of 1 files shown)


elthariel:
I don't think the DCO is required for documentation change, no ?

stuartpreston:
+1

elthariel:
cc @jkeiser @tyler-ball

PR #572, opened 2017-04-12
Fixes for chef13 and cheffish5.x/cheffish-13.x

jjlimepoint:
* driver_for_url needs a to_hash added to its config for the string url case (like used with vsphere, for exmple), since MergedConfig only takes hashes, but config is a mash

* resource creation functions no longer take caller

The first of these is actually required for chef12 with cheffish5 as well, the other two commits are chef-13 compatability


tyler-ball:
Thanks for this PR!

PR #575, opened 2017-04-17
Update README.md

jjasghar:
Fixed chart, added Chef-Partners maintaining.


jjasghar:
Do'h i guess README through Github editor doesn't count for DCO 😢 🐼

PR #576, opened 2017-04-17
hash hash baby.

jjasghar:
It seems there is a bug with how the mash is created with more...full
Chef::Config options. This forces the `config` to be a hash to make sure
that it is created correctly for Cheffish 5+.

Signed-off-by: JJ Asghar <jj@chef.io>


tyler-ball:
Closing this in favor of #572

PR #578, opened 2017-04-21
Putting back in place some development dependencies I removed

tyler-ball:
@lamont-granquist I thought I was being smart, but I think I just did not understand what you were doing in #577 and #568 

Am I correct in thinking that you have loosened these dependencies so they will work with upcoming Chef and ChefDK releases?


lamont-granquist:
chef 13 / ohai 13 have been released, so you shouldn't need to pin to git master.

Issue #285, opened 2015-02-18
set_attributes and get_attributes override for WindowsMachine


joelwilson:
Using the machine_file resource to upload a file to a windows machine results in an error:

```
set_attributes not overridden on Chef::Provisioning::Machine::WindowsMachine
```

The cause is pretty obvious; [the `set_attributes` and `get_attributes` in WindowsMachine.rb are not implemented](https://github.com/chef/chef-provisioning/blob/master/lib/chef/provisioning/machine/windows_machine.rb#L78-L84) in the windows machine provider.

Reproduce:

``` ruby
machine_file 'C:/chef/trusted_certs/chefserv01_companyname_com.crt' do
  local_path "#{Chef::Config[:trusted_certs_dir]}/chefserv01_companyname_com.crt"
  machine HOSTNAME
end
```

Full error message

```
- upload file U:/.chef/trusted_certs/chefserv01_companyname_com.crt to C:/chef/trusted_certs/chefserv01_castingn
etworks_com.crt on wintest01
================================================================================
Error executing action `upload` on resource 
'machine_file[C:/chef/trusted_certs/chefserv01_companyname_com.crt]'

================================================================================

RuntimeError
------------
set_attributes not overridden on Chef::Provisioning::Machine::WindowsMachine

Resource Declaration:
---------------------
# In C:/Users/auser/Code/chef/provision-windows-test/recipes/win-test.rb

  7: machine_file 'C:/chef/trusted_certs/chefserv01_companyname_com.crt' do
  8:   local_path "#{Chef::Config[:trusted_certs_dir]}/chefserv01_companyname_com.crt"
  9:   machine HOSTNAME
 10: end
 11:

Compiled Resource:
------------------
# Declared in C:/Users/auser/Code/chef/provision-windows-test/recipes/win-test.rb:7:in `from_file'

machine_file("C:/chef/trusted_certs/chefserv01_companyname_com.crt") do
  action :upload
  retries 0
  retry_delay 2
  guard_interpreter :default
  chef_server {:chef_server_url=>"https://chefserv01.companyname.com/organizations/xxxxxx", 
:options=>{:client_name=>"auser", :signing_key_filename=>"U:\\.chef/client.pem"}}
  cookbook_name "@recipe_files"
  recipe_name "C:/Users/auser/Code/chef/provision-windows-test/recipes/win-test.rb"
  local_path "U:/.chef/trusted_certs/chefserv01_companyname_com.crt"
[...]

PR #582, opened 2017-05-17
Ignore group/owner/mode attrs on Windows machine until implemented

paustin01:
Implementing @joelwilson 's fix from https://github.com/chef/chef-provisioning/issues/285.

Attempting to set group/owner/mode on uploaded files bombs on Windows nodes with the message 

`set_attributes not overridden on Chef::Provisioning::Machine::WindowsMachine`

We should ignore this on windows nodes until implemented.


tyler-ball:
I'll add a comment around that block in case we ever get time to fix the `WindowsMachine`

PR #583, opened 2017-05-25
We can’t forward a socketless Zero server, so show a nice(ish) error

coderanger:
Because of https://github.com/chef/chef/pull/6177


tyler-ball:
Thanks @coderanger !

Issue #587, opened 2017-07-24
401 error on node \ client  precreation


gregsher88:
I am receiving a 401 error when trying to precreate chef objects as part of a chef-provisioning run. 

I have followed instructions for using knife-acl to define a provisioners group as described in the chef-provisioning readme, however the issue persists.

I am able to create nodes successfully with knife using the same node_name and client_key.

A gist of the code being used to provision is here: [(https://gist.github.com/gregsher88/3a3bcb819db812eb48aabb415d109adb)](https://gist.github.com/gregsher88/3a3bcb819db812eb48aabb415d109adb)

```
Recipe: (chef-apply cookbook)::(chef-apply recipe)
  * machine[2elastic-dev] action converge

    ================================================================================
    Error executing action `converge` on resource 'machine[2elastic-dev]'
    ================================================================================

    Net::HTTPServerException
    ------------------------
    401 "Unauthorized"

    Resource Declaration:
    ---------------------
    # In c:\code\chef-provisioning-hyperv\examples\elastic-cluster-dev.rb

      6: machine "2elastic-dev" do
      7:
      8:  chef_server :chef_server_url => "https://command-chef-live.cloudapp.net/organizations/command",
      9:   :client_name =>   "#{Chef::Config[:node_name]}".downcase,
     10:   :signing_key_filename =>  Chef::Config[:client_key]
     11:
     12:     machine_options hyperv_options: {
     13:       path: '\\\\storage01-hv01\\TLIQA\\VM_STORAGE\\VMQAHV2K\\',
     14:       vhd_path: "\\\\storage01-hv01\\TLIQA\\VM_STORAGE\\VMQAHV2K\\2elastic-dev.vhdx",
     15:       switch_name: 'DEV',
     16:       generation: '2',
     17:       cpus: '2',
     18:       vlan: '2510',
     19:       vm_guest: 'linux',
     22:       network: {
     23:         ip: "172.25.11.121",
     24:         subnet: '255.255.254.0',
     25:         gateway: '172.25.10.1',
     26:         dns: %w(172.25.10.254 172.25.10.7)
     27:       }
     28:     },
     29:                     convergence_options: {
     30:                       chef_config: 'ssl_verify_mode :verify_none' # String containing additional text to inject
 into client.rb
     31:                     }
     32: end

    Compiled Resource:
    ------------------
    # Declared in c:\code\chef-provisioning-hyperv\examples\elastic-cluster-dev.rb:6:in `run_chef_recipe'

[...]

PR #588, opened 2017-07-28
Add options hash to 'with_chef_server' example 

gregsher88:
This is the resolution to issue #587 "401 error on node \ client precreation"


tas50:
Thanks for the fix

PR #596, opened 2017-09-08
Add an option to never use SSH port forwardin

jjlimepoint:
We have situations where our templates have SSH port forwrding hard disabled - this patch allows the specifying of an SSH option to never even try.  It should alsop help with the "chef_server_url localhost" issue that crops up from time to time.


tyler-ball:
Thanks @jjlimepoint

PR #602, opened 2018-03-02
chef_config really should take a hash of options rather than a string

jjlimepoint:
This PR make it optionally do so.  Even worked once for me!


jjlimepoint:
The travis failure seems to be chef14 conflicts rather than patch problems - is there a way to poke it to re-run?

jjlimepoint:
This is now significantly more tested, huzzah

PR #606, opened 2018-03-12
Add Errno::ECONNABORTED to the list of rescued errors in the available? function

### LINKED ISSUE #605: winrm available? function exception when using SSL
The winrm available? function will throw an Exception (SSL_Connect) when attempting to use SSL if the IP is available but the HTTPS winrm endpoint does not yet exist on the system.

For example, using test-kitchen to test a windows node using chef-provisioning-vsphere will result in the following exception when the VM is up, but an HTTPS endpoint is not yet available (In my case, an SSL cert is generated and the winrm HTTPS transport configured as part of the VM customization.  While sysprep is being run, there is no HTTPS listener and the following exception occurs once the IP is discovered):

```
>>>>>> ------Exception------
-
>>>>>> Class: Kitchen::ActionFailed
>>>>>> Message: 1 actions failed.
>>>>>>     Failed to complete #create action: [An established connection was aborted by the software in your host machine. - SSL_connect] on platform-suite
>>>>>> ----------------------
>>>>>> Please see .kitchen/logs/kitchen.log for more details
>>>>>> Also try running `kitchen diagnose --all` for configuration
```

Adding Errno::ECONNABORTED to the list of rescued exceptions allows the process to function as it normally would with a standard HTTP listener.

lomeroe:
Attempts to attach over winrm using SSL when the SSL endpoint does not
yet exist will result in an ECONNABORTED exception.

Obvious fix.

Fixes #605

### DIFF
--- a/lib/chef/provisioning/transport/winrm.rb
+++ b/lib/chef/provisioning/transport/winrm.rb
@@ -79,7 +79,7 @@ module Provisioning
         Chef::Log.debug("unavailable: winrm authentication error: #{$!.inspect} ")
         disconnect
         false
-      rescue Timeout::Error, Errno::EHOSTUNREACH, Errno::ETIMEDOUT, Errno::ECONNREFUSED, Errno::ECONNRESET, ::WinRM::WinRMError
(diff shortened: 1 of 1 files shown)


tyler-ball:
Thanks @lomeroe !

PR #609, opened 2018-04-05
allow cheffish14, since it does not break us

jjlimepoint:
Since chef14 can be used safely, allow it


jjlimepoint:
Caveat: for some reason, I've occasionally gotten a chef14 with an old cheffish with a bug, where mergedconfig was just broken - i'm pretty sure this is a packaging problem on the chef14 side, though, because it's been correct in cheffish's git for 5 months now.  I didn't get that bad chef14 today though... YMMV!

eheydrick:
This can be closed since https://github.com/chef/chef-provisioning/pull/611 bumped cheffish.

PR #615, opened 2018-07-18
Ensure that machines are disconnected after machine_batch actions

horazont:
TL;DR: Make sure that machine_batch disconnects the transport to avoid broken port forwardings for chef-zero.

#### Motivation / Bug

In our setup, we have a cookbook which uses Chef Provision to converge the target nodes. We invoke that cookbook with chef-client and we use chef-zero. We are using the ssh driver. In our setup, we also converge the same target node more than once during a single Chef Provision chef-client run.

If not disconnected, the ssh transport keeps its connection open, but in a weird state: The port forwarding for the chef zero server (port 8889 by default) is somewhat broken; trying to make a curl against it leads to a timeout and the request is never seen by the chef-zero server on the provisioning node. The next converge action then sees the existing port forwarding (which is broken, however) and chooses a new random port for its own port forwarding. The chef-client on the target node does *not* get informed about the changed chef server URL (which may be a separate issue) and tries to use localhost:8889. As this is the old, broken port forwarding, this leads to the chef-client run to never even really start.

(I also wrote about our [journey towards this bugfix in the slack channel](https://chefcommunity.slack.com/archives/C1W2WL80P/p1531905486000270).)

#### The Fix

The disconnect method of the ssh transport is meant to close the session. It also logs a debug message when it is invoked. By looking through the debug logs, we found that it is *not* invoked in our case.

Digging deeper into things, we found that machine_batch never invokes the disconnect method of the machine and thus the transport. This PR attempts to fix that.

#### Tests

I was unable to find tests for the ``machine_batch`` resource. If there are any, let me know and I’ll happily add a new test/adapt the existing tests to test for this bug and fix.

Applying this patch solves the issue I described above and does not re-introduce any new issues in our case.

### DIFF
--- a/lib/chef/provider/machine_batch.rb
+++ b/lib/chef/provider/machine_batch.rb
@@ -41,29 +41,37 @@ class MachineBatch < Chef::Provider::LWRPBase
 
   action :setup do
     with_ready_machines do |m|
-      m[:machine].setup_convergence(m[:action_handler])
-      m[:spec].save(m[:action_handler])
-      Chef::Provider::Machine.upload_files(m[:action_handler], m[:machine], m[:files])
+      begin
+        m[:machine].setup_convergence(m[:action_handler])
+        m[:spec].save(m[:action_handler])
+        Chef::Provider::Machine.upload_files(m[:action_handler], m[:machine], m[:files])
+      ensure
+        m[:machine].disconnect
+      end
     end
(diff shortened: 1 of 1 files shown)


horazont:
I added the Signed-off line I missed before. Can we get this merged, or are there any issues with it?

